PR #12, opened 2024-08-11
Add Dilated Sliding Window mask_mod

### LINKED ISSUE #2: Add Dilated Sliding Window mask_mod
# Summary
Paper: https://arxiv.org/pdf/2004.05150
LongFormer

sangyeon-k:
# Summary
- Resolves #2 
- Adds mask_mod for a dilated sliding window with causal masking

# Visualization
<p align="center">
  
</p>

### DIFF
--- a/attn_gym/masks/dilated_sliding_window.py
+++ b/attn_gym/masks/dilated_sliding_window.py
@@ -1,0 +1,60 @@
+import torch
+from torch.nn.attention.flex_attention import _mask_mod_signature, and_masks
+from attn_gym.masks import causal_mask
+
+
+def generate_dilated_sliding_window(window_size: int, dilation: int) -> _mask_mod_signature:
+    """Generates a dilated sliding window attention mask.
+    Args:
+        window_size: The size of the sliding window.
+        dilation: The dilation factor for the sliding window.
+
+    Note:
+        We assume that the window size represents the lookback size and we mask out all future tokens
+        similar to causal masking.
+    """
+
+    def dilated_sliding_window(b, h, q_idx, kv_idx):
+        diff = q_idx - kv_idx
+        in_window = (diff >= 0) & (diff < window_size * dilation)
+        is_dilated = (diff % dilation) == 0
+        return in_window & is_dilated
+
+    dilated_sliding_window_mask = and_masks(dilated_sliding_window, causal_mask)
+    dilated_sliding_window_mask.__name__ = (
+        f"dilated_sliding_window_{window_size}_dilation_{dilation}"
+    )
+    return dilated_sliding_window_mask
+
+
+def main(device: str = "cpu"):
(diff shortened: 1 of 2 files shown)


attn_gym/masks/dilated_sliding_window.py
```diff
@@ -0,0 +1,60 @@
...
+        dilation: The dilation factor for the sliding window.
+
+    Note:
+        We assume that the window size represents the lookback size and we mask out all future tokens
+        similar to causal masking.
+    """
+
+    def dilated_sliding_window(b, h, q_idx, kv_idx):
+        diff = q_idx - kv_idx
+        in_window = (diff >= 0) & (diff < window_size * dilation)
```
drisspg, line 19: hmm from the paper its not clear to me that its always causal

what about torch.abs(diff) < window_size ?

drisspg: One other nit I think that its clearer if we keep the window_size and dilation separate

e.g. to recreate the paper (if we didnt have the and_causal mask)
we would set window_size = 8 and dilation = 2

drisspg (commented):
Thanks for opening up the PR! Left two comments/ questions let me know what you think

attn_gym/masks/dilated_sliding_window.py
sngkim: > what about torch.abs(diff) < window_size ?

I thought it would be good to make this implementation consistent with `attn_gym/masks/sliding_window.py`.

However, seems reasonable to follow the non-causal way the paper described. I will update the `generate_dilated_sliding_window()` function.

sngkim: > One other nit I think that its clearer if we keep the window_size and dilation separate
> 
> e.g. to recreate the paper (if we didnt have the and_causal mask)
> we would set window_size = 8 and dilation = 2

Maybe, I missed something. Can you please explain what does it mean by "if we keep the window_size and dilation separate"?
Did you mean setting window_size = 8 and dilation = 2?

sngkim:
Hi @drisspg, thanks for the review!
I left a follow-up question for clarification. After that, I will update this PR!

attn_gym/masks/dilated_sliding_window.py
drisspg: Ohh I just meant that the dilation factor doesnt have any impact on the absolute size of the window. 
`window_size * dilation` -> `window_size`

So the "potential" size of the window is 16 elements (8 forward, 8 backward ) but a dilation factor knocks out half and we end up up with 4 on both sides. We dont extend the window so as to capture more elements

drisspg:
Landed here: https://github.com/pytorch-labs/attention-gym/pull/85

PR #16, opened 2024-08-14
NATTEN example

Birch-san:
- renamed existing "NATTEN" example to [SASA](https://arxiv.org/abs/1906.05909). see [Ali Hassani's explanation & visualization](https://youtu.be/Ya4BfioxIHA?si=0aPdrXG5q3P_RcQr&t=92).
- added NATTEN mask_mod to masks directory ([repo](https://github.com/SHI-Labs/NATTEN))

See my [parity test](https://github.com/Birch-san/natten-fwd-ad/blob/b7f812619fe667b4a1120060087c9f8ce17b598c/script/demo_flex.py) which demonstrates that the FlexAttention implementation is allclose equivalent to NATTEN and also to masked SDPA.

3x3 NATTEN kernel on 6x6 canvas  
natten_c6x6_k3x3


drisspg (approved):
Awesome! thanks for contributing and fixing the previous example

PR #48, opened 2024-09-29
Add explicit statement in README.md for installing the required version of PyTorch

### LINKED ISSUE #47: Add explicit code statement in README for installing PyTorch 2.5 (currently only available as nightly)
Following the suggested installation instructions in the README installs only the latest stable release of Torch (currently 2.4.1) but the package requires version 2.5.0 or above.

A code line should be added in the README to explicitly specify the step of installing the Nightly release of PyTorch (nightly only for now, can be updated later).

```bash
pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu124
```

Otherwise, the stable PyTorch version is installed and leads to import errors because of file name differences (`_flex_attention.py` vs `flex_attention.py`)

sachinkadyan7:
Fixes #47 

Installing the (nightly) 2.5.0 version of PyTorch first before the other dependencies are installed. Otherwise, the current stable version of PyTorch (2.4.1) is installed by default.

To avoid that scenario, I am adding an explicit `pip` statement to install the PyTorch 2.5.0 nightly version first before anything else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ This repository aims to provide a playground for experimenting with various atte
 ```bash
 git clone https://github.com/pytorch-labs/attention-gym.git
 cd attention-gym
+pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu124
 pip install .
 ```
 


drisspg:
Left a comment on the issue, but we already have this in warning. We can bump the toml depedency to be >2..5 once it this version is officially released. The RC is in process of being pushed

PR #100, opened 2025-01-03
Reduce memory usage

oraluben:
by run attn impls one by one and clear CUDA memory after each case.

This allows to run benchmark on nvidia A10.


drisspg (approved):
Looks good, thanks

PR #128, opened 2025-03-18
Fix natten x y dims

fcossio:


drisspg (approved):
Looks good, thanks for the fix

PR #150, opened 2025-07-11
Fix bench_block_mask so it can run.

lucasb-eyer:
I fixed almost all issues I encountered while trying to run the benchmark. One thing I'm still seeing and know nothing about, is this error for some of the configs:

```
Failed to run config ExperimentConfig(B=8, H=16, M=8192, N=8192, mask_mod_name='doc_mask_mod'):
('Failed to divide all profiling events into #repeat groups. #CUDA events: %d, #repeats: %s', 65, 2)
```

### DIFF
--- a/attn_gym/masks/document_mask.py
+++ b/attn_gym/masks/document_mask.py
@@ -1,5 +1,6 @@
 """Generates a document causal attention mask based on a document ID tensor"""
 
+import random
 from typing import List, Union
 
 import torch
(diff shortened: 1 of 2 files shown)


drisspg:
Ahhh that issue has been popping up more and more and its annoying. do_bench_with_profiling is the profiling function that produces the closest timings to what NCU will report which is why I like it. However I think there are new kineto events that aren't handled correctly. 

Solution is to just swap w/ tritons do_bench

drisspg:
No worries, also, I fixed a few more things here: https://github.com/pytorch-labs/attention-gym/pull/153

PR #202, opened 2026-05-05
Add exclusive self-attention(XSA) output modifier

ishaannk:
Introduces exclusive_output_mod() - a post-output modifier that removes
the self-value projection from attention output, guaranteeing Z ⊥ V.

This is the first output_mod pattern in attention-gym. Existing score_mods
cannot implement XSA because it operates on the weighted-sum output, not
pre-softmax scores.

### DIFF
--- a/attn_gym/mods/exclusive_sa.py
+++ b/attn_gym/mods/exclusive_sa.py
@@ -1,18 +1,8 @@
 from __future__ import annotations
-
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
 from torch import Tensor
-
-try:
-    from torch.nn.attention.flex_attention import flex_attention
-
-    HAS_FLEX_ATTENTION = True
-except ImportError:
-    HAS_FLEX_ATTENTION = False
-
-
+from torch.nn.attention.flex_attention import flex_attention
 def exclusive_output_mod(Y: Tensor, V: Tensor) -> Tensor:
     """Remove each attention output vector's projection onto its value vector."""
     v_sq_norm = (V * V).sum(dim=-1, keepdim=True)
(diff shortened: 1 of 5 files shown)


attn_gym/mods/exclusive_sa.py
```diff
@@ -0,0 +1,111 @@
...
+        x: Tensor,
+        score_mod=None,
+        block_mask=None,
+        is_causal: bool = True,
+    ) -> Tensor:
+        Q = self._split_heads(self.W_q(x))
+        K = self._split_heads(self.W_k(x))
+        V = self._split_heads(self.W_v(x))
+
+        if HAS_FLEX_ATTENTION and (score_mod is not None or block_mask is not None):
```
drisspg: we can assume flex-attention is always avail

attn_gym/mods/latent_attention.py
```diff
@@ -15,16 +11,16 @@ def generate_mla_rope_score_mod(
     num_heads: int,
     scale: float = 1.0,
 ) -> _score_mod_signature:
-    """Returns an MLA RoPE score modification function to be used w/ FlexAttention
+    """Return an MLA RoPE score modification function for FlexAttention.
```
drisspg: tell claude to not touch this

ishaannk: Claude is sleeping! I had done it Please review again

ishaannk:
Hey @drisspg something still wrong?

examples/xsa_attention.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def example_composable():
+    print("=" * 60)
+    print("Example 2: XSA + ALiBi score_mod")
+    print("=" * 60)
+
+    Q, K, V = make_qkv(T=128)
+
+    if HAS_FLEX:
```
drisspg: same we shoudl be able to assert we have flex

attn_gym/mods/exclusive_sa.py
```diff
@@ -0,0 +1,83 @@
...
+        cleanup_tol = torch.finfo(Z.dtype).eps * Z.shape[-1] * reference_norm
+        Z = torch.where(
+            (v_sq_norm > 0) & (residual_norm <= cleanup_tol),
+            torch.zeros_like(Z),
+            Z,
+        )
+    return Z
+
+
+class XSAMultiheadAttention(nn.Module):
```
drisspg: I acutally dont want to be pesrscriptve on nn.Module thi sis fine touse int eh examples but I don think we should ship this by default

ishaannk (commented):
Hey done @drisspg please check and thanks for addressing i am a kid

drisspg:
Hey sorry looked at this again; I dont like it its too opinionated and I dont think it provides much value over users just writing their own suffix modifiers in their own nn moduels

PR #225, opened 2026-07-22
Add Compressed Sparse Attention

floatingtrees:
# Compressed Sparse Attention

Compressed sparse attention is a somewhat recently introduced variant of attention introduced in [DeepSeek v4](https://arxiv.org/html/2606.19348v1). 

Mathematically, it can be summarized as: 

1. Compute `Q_latent`, `KV`, `C_a`, `C_b`, `Z_a`, `Z_b`, `K_Ia`, `K_Ib`, `Z_Ia`, and `Z_Ib` by projecting out from the residual stream with learned weight matrices

2. Normalize `Q_latent`, compute `Q`, `Q_I` by projecting `Q_latent` with learned weight matrices

3. Create blocks as follows: 

   a. Shift `Z_b`, `C_b` by `block_size` tokens to the right along the sequence dimension; truncate anything that goes beyond `sequence_length`, and pad the left appropriately. 

   b. Reshape tensors from:

   ```text
   B, H, S, D
   ```

   to:

   ```text
   B, H, S // block_size, block_size, D
   ```

   c. Compute:

   ```text
   P = softmax(cat([Z_a + B_a, Z_b + B_b]))
   ```

   with all operations performed along the `block_size` dimension

   d. Split `P` in half along the `block_size` dimension to create `S_a`, `S_b`

   e. Return:

   ```text
   sum(C_a * S_a + C_b * S_b, reduce over block_size dimension)
   ```

4. Create:

   ```text
   compressed_KV = compress(C_a, C_b, Z_a, Z_b, B_a, B_b)
   compressed_indices = compress(K_Ia, K_Ib, Z_Ia, Z_Ib, B_Ia, B_Ib)
   ```

   where `B_x` are learnable biases

5. Apply YaRN rope to vectors along selected dimensions and normalize

6. Score each `compressed_index` block:

   a. 

   ```text
   per_head_score = relu(Q_I @ compressed_indices.T)
   ```

   reduce across `head_dim`

   b. 

   ```text
   block_score = W_I * per_head_score
   ```

   reduce across `num_heads`

7. For each query, select blocks with top `K` scores and `KV` corresponding to the last `sliding_window_size` tokens, call that `KV_final`

8. Perform:

   ```text
   softmax(Q @ KV_final.T / sqrt(head_dim)) @ KV_final
   ```

# This PR contains

1. A batch invariant reference implementation

2. A triton and CuTe DSL backend 

3. Tests covering forward and backward implementation for all backends

The CuTe DSL backend in integrated into [this](https://github.com/floatingtrees/torchtitan/tree/dev/attn-gym-csa) torchtitan branch for DeepSeek v4, and the triton backend is integrated into [this](https://github.com/floatingtrees/torchtitan/tree/dev/attention-gym-triton) torchtitan branch for the same model. 

# Public API

```python
from attn_gym.sparse import compressed_sparse_attention
compressed_sparse_attention(*args)
```

# To run either integration

```bash
pip install -r requirements.txt
pip install git+https://github.com/floatingtrees/attention-gym.git@705e5e7aa765436f47ae237fde257f98426864df
NGPU=4 MODULE=deepseek_v4 CONFIG=deepseek_v4_debugmodel ./run_train.sh     --override.imports torchtitan.models.deepseek_v4.attention_gym_csa  --training.dtype bfloat16     --training.steps 3000     --training.local_batch_size 4     --training.seq_len 4096
```

# Results for each integration on 4xGB200s

## Original baseline

`210 ms/step` at `head_dim=256`, `80 TFLOPs`, loss drops from `8` to `0.72` by step `3000`. 

## CuTe DSL backend

`199ms/step` at `head_dim=512`, `150 TFLOPs`, loss drops from `8` to `0.7` by step `3000`. (The CuTe DSL backend does not support `head_dim != 512`)

## Triton backend

`150ms/step` at `head_dim = 256`, `90 TFLOPs`, loss drops from `8` to `0.69` by step `3000`. 

# Compatibilities
| Compatibility | Reference | Triton | CuTe DSL |
|---|---:|---:|---:|
| Forward | Yes | Yes | Yes |
| Backward | Yes | Yes | Yes |
| Training| Yes | Yes | Yes |
| Decode | No | No | No |
| Packed sequences | No | No | No |
| `torch.compile` | Yes | Yes | Yes |
| GPU architecture | CPU/CUDA | CUDA | `SM100` |
| `Head dim` | Any | `<=256` | `512` |
| Shared KV required | No | No | Yes |
| Batch-invariance guarantee | Exact | 3e-2 | 3e-2 |

# Limitations

The CuTe DSL backend only supports `SM100`, `share_kv = True`, and `head_dim = 512`. `head_dim = 512` is what DeepSeek v4 uses. A kernel for `head_dim=512` differs substantially to kernels with `head_dim <= 256` due to shared memory constraints. Use cases that don't fit this can use the triton backend. 

The kernels were generated with Codex


drisspg:
give me a sec ill take a crack at reviewing this

floatingtrees:
Hi Driss, the end to end training run numbers for CuTe DSL are slightly outdated. For the CuTe DSL backend, I was able to hit 1.4 PFLOPs/GPU on [this torchtitan branch](https://github.com/floatingtrees/torchtitan/tree/dev/attn-gym-csa) on 4xGB200s with this command: 

```
 NGPU=4 MODULE=deepseek_v4 CONFIG=deepseek_large_debug_config ./run_train.sh
```

drisspg:
@floatingtrees awesome, do we have micro benchmarks as well for the kernels?

floatingtrees:
There are benchmarks for both backends in the `benchmarks/sparse/benchmark_compressed_sparse_attention` folders that measure TFLOPs for each backend. The CuTe DSL backend hits roughly 500 TFLOPs.

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+
+import torch
+
+Backend = Literal["eager", "triton", "cute"]
+Mode = Literal["auto", "chunked", "recurrent"]
+
+_CUTE_RUNTIME_DEPENDENCIES = (
+    ("cuda-python", "cuda.bindings.driver", "13.3.1"),
+    ("nvidia-cutlass-dsl", "cutlass.cute", "4.5.2"),
+    ("flash-attn-4", "flash_attn.cute.interface", "4.0.0b17"),
```
drisspg: why b417?

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+    ("quack-kernels", "quack", "0.5.0"),
```
drisspg: this is a lil odl as well

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+from importlib import metadata
+import math
+from typing import Literal
+
+import torch
+
+Backend = Literal["eager", "triton", "cute"]
+Mode = Literal["auto", "chunked", "recurrent"]
+
+_CUTE_RUNTIME_DEPENDENCIES = (
```
drisspg: actuqlly can we just add an extra install to the py proejct.toml for [sparse]

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+        "1.25.0",
+    ),
+)
+
+_cute_implementation: Callable[..., torch.Tensor] | None = None
+_cute_initialization_error: Exception | None = None
+
+
+def _validate_inputs(
+    tensors: tuple[tuple[str, object], ...],
```
drisspg: one nit is that claude really liekes this `_thing` behaviro I lwo key hate it I think its okay fto ahve at , so specifally at least for thse lets inline these typs or jstu really anme its fine if it _

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+
+
+def _load_eager_implementation() -> Callable[..., torch.Tensor]:
+    from . import reference
+
+    # The checked-in oracle uses ``math`` in RoPE without importing it. Keep the reference file
+    # untouched while making its public eager entry point runnable.
+    reference.math = math
+    return reference.CSA
```
drisspg: hmm this is gross let sjsut import it

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def _load_triton_implementation() -> Callable[..., torch.Tensor]:
+    try:
+        from .triton import compressed_sparse_attention as implementation
+    except ModuleNotFoundError as error:
+        if error.name in (f"{__package__}.triton", "triton"):
+            raise RuntimeError(
+                "The Triton backend for compressed sparse attention is not available; "
```
drisspg: we can assume Triton is installed since we shipped it with PyTorch.

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+    return implementation
+
+
+def _validate_cute_dependencies() -> None:
+    """Validate the pinned CuTe stack during backend initialization."""
+    problems = []
+    for distribution, module, expected_version in _CUTE_RUNTIME_DEPENDENCIES:
+        try:
+            importlib.import_module(module.partition(".")[0])
+            importlib.import_module(module)
```
drisspg: Here we just add the PI project dependencies, and this can simplify a lot of this stuff.

attn_gym/sparse/compressed_sparse_attention/api.py
```diff
@@ -0,0 +1,388 @@
...
+    elif backend == "triton":
+        implementation = _load_triton_implementation()
+    else:
+        if _cute_implementation is None:
+            assert _cute_initialization_error is not None
+            raise RuntimeError(str(_cute_initialization_error)) from _cute_initialization_error
+        implementation = _cute_implementation
+
+    return implementation(
+        Q,
```
drisspg: Let's use match case here. We're on a new enough Python. Also, yeah, same here. We can assume the implementations are there.

attn_gym/sparse/compressed_sparse_attention/cute/__init__.py
```diff
@@ -0,0 +1,1848 @@
...
+The public entry point owns the complete CSA data path: compression, RMS normalization,
+RoPE, index scoring/top-k, local and compressed attention, learned-sink merge, and inverse
+RoPE.  PyTorch is used only for validation, allocation, and cached RoPE constants; input-
+dependent tensor arithmetic is launched through CuTe DSL kernels.
+"""
+
+from __future__ import annotations
+
+from collections import OrderedDict
+import math
```
drisspg, line 12: This is too big. I don't think we wanna ship all this.  Can you explain to me what kernels we are adding and where and how they compose? This is a little too much. I want to figure out what's going on. Want to be a little bit less prescriptive on how to glue these pieces together and more provide implementations.

floatingtrees:
Hi Driss, thanks for the feedback. I got rid of add the dependency checks in API and added optional dependencies for sparse. The CuTeDSL backend has the following steps: 

Preprocessing kernels: These all do some combination of compress + RoPE + RMS norm. There are couple different kernels for this: optimized ones for specific head shapes, a kernel that doesn't perform compression (for the sliding window branch), etc. 

Indexing kernels:
1. Score computation: There's one optimized kernel for this; for shapes that haven't been targeted for this, there's a padding and less kernel as well. There's also another kernel that ensures deterministic tiebreak behavior. 
2. Block Selection Kernel: Calls a cuDNN kernel to select the `topk` blocks. 

Attention: Calls a vendored and slightly modified FA4 kernel. 

Epilogue: Calls a kernel to reverse rope and another kernel to fix numerical accuracy issues. 

Backwards: For certain shapes, the backward pass calls a vendored kernel from cuDNN. Otherwise, there's a fallback path that's quite similar structurally to the forward pass. 

I think there are a couple of kernels I can trim out without hurting performance too much for the shapes we care about. Do you have any thoughts about what we should do, or if we should expose more of this to the user?

attn_gym/sparse/compressed_sparse_attention/api.py
floatingtrees: fixed

drisspg:
@floatingtrees give me a sec still doing some understanding of the overall structure will provide feedback in an hour

drisspg:
@floatingtrees

Okay overall this pr is to hard to review and land as is and we need to break up into smaller chunks;

for instance it seems like the triton path uses much more eager code and is much smaller/ more manageable to review and can act as the base line. Lets make a seperate PR for that. Then lets addinng benchmark scripts. And we can slowly incorporate more and more cutedsl  / friends and very concretly show the speed ups over the existing baseline.

One more meta note is that I need to understand better why we need to fork FA4. And want that part in a seperate PR so that I can more fully grok it.

floatingtrees:
Ok, that makes sense. I'll group the reference and triton implementation into one PR and add it tomorrow.

floatingtrees:
@drisspg, I opened the first sub PR in #232.

floatingtrees:
Since we already merged selected attention, I think this PR can be closed.

PR #232, opened 2026-07-30
Adding Reference and Triton implementations for Compressed Sparse Attention

floatingtrees:
First sub-PR from #225. This PR adds the reference and Triton implementations for CSA, along with correctness tests for the Triton implementation.


attn_gym/sparse/compressed_sparse_attention/reference.py
```diff
@@ -0,0 +1,280 @@
...
+    """Split the sequence dimension into block and within-block dimensions."""
+    return x.reshape(
+        *x.shape[:-2],
+        x.shape[-2] // compression_rate,
+        compression_rate,
+        x.shape[-1],
+    )
+
+
+def compress(C_a, C_b, Z_a, Z_b, B_a, B_b, compression_rate):
```
drisspg, line 61: these refernce fluciotns should have google style doc blocks

attn_gym/sparse/compressed_sparse_attention/reference.py
```diff
@@ -0,0 +1,280 @@
...
+    x: torch.Tensor,
+    positions=None,
+    base: float = 160_000.0,
+    original_seq_len: int = 65_536,
+    factor: float = 16.0,
+    beta_fast: float = 32.0,
+    beta_slow: float = 1.0,
+    position_offset: int = 0,
+    inverse: bool = False,
+) -> torch.Tensor:
```
drisspg, line 96: ditto all these funcs since they are the reference to where they come from either the paper or no jsut an explanation;

drisspg (changes requested):
I don’t think compressed_sparse_attention should be the main API. It is way too big, does way too many things, and will be a real pain to maintain and extend. Right now, the public signature combines a lot that I would prefer build up from primitives or examples we provide. I don’t think we should commit to that entire composition as the abstraction Attention Gym provides.                                                                                                                                                                     
                                                                                                                                                                                                                                        
Instead, we should identify the smallest primitives for which users actually need to leave ordinary PyTorch and need specialized implementations for performance. Everything else should remain ordinary PyTorch composition unless we have a concrete reason to expose or optimize it.                                                                                                                                                                                       
                                                                                                                                                                                                                                        
 I need to look more closely at whether `_launch_selected_attention` is exactly the right boundary, but the general split makes sense to me. 

The compressed attention kereel would take a prepared query, local KV, compressed KV, and selected block indices, and jointly compute the local-window and selected-block attention. The shared normalization and attention-sink behavior would also belong to that operation, since computing those paths separately would not preserve the intended softmax semantics. This appears to be the part that actually needs a specialized implementation and custom backward.                                                                                                          
                                                                                                                                                                                                                                        
 One question we need to answer is how users are expected to prepare those inputs in a semantically clear way. In particular:                                                                                                           
                                                                                                                                                                                                                                        
 - How is the query prepared?                                                                                                                                                                                                           
 - How are the local and compressed KV tensors prepared?                                                                                                                                                                                
 - How are the selected block indices produced?                                                                                                                                                                                         
 - Which normalization and positional encoding conventions are required?                                                                                                                                                                
 - Do any of these preparation steps require custom kernels for performance?                                                                                                                                                            
                                                                                                                                                                                                                                        
 If those steps can be expressed clearly and efficiently using ordinary PyTorch, that is completely fine. We should not add APIs or custom kernels for them merely for completeness. A reference implementation should instead show exactly how to prepare the inputs and document the required semantics.                                                                                                                                                                 
                                                                                                                                                                                                                                        
 Very concretely, I would like us to provide a well-documented example that composes these pieces into a complete, opinionated compressed-attention module. That example can implement the full recipe currently contained in `compressed_sparse_attention`: the projections and gates, interleaved compression, RMSNorm, RoPE, index scoring, causal filtering, top-k selection, the selected-attention primitive, and any required output transformation. It can have a larger and more model-specific interface because its purpose is to demonstrate how to build the complete module, not to establish a stable library abstraction. And this should be well documented and show users how to it can be done.                                                 
                                                                                                                                                                                                                                        
 The example should be tested end to end so that we know composing the smaller pieces reproduces the current reference implementation. Ideally, it should also document the tensor shapes and the semantic contract between preparation and the optimized primitive. That gives users something concrete they can copy or adapt without forcing us to make every intermediate helper part of the public API.                                                                   
                                                                                                                                                                                                                                        
 For the public API, I think we should start with only the minimal operation or set of operations that users need to build efficient compressed attention. A possible shape would be something along these lines:                       
                                                                                                                                                                                                                                        
 ```python                                                                                                                                                                                                                              
   selected_attention(                                                                                                                                                                                                                  
       query,                                                                                                                                                                                                                           
       compressed_kv,                                                                                                                                                                                                                   
       local_kv,                                                                                                                                                                                                                        
       selected_blocks,                                                                                                                                                                                                                 
       attention_sink=...,                                                                                                                                                                                                              
       compression_rate=...,                                                                                                                                                                                                            
       sliding_window_size=...,                                                                                                                                                                                                         
       backend=...,                                                                                                                                                                                                                     
   )                                                                                                                                                                                                                                    
 ```                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                        
 I am not attached to this exact name or signature.. The important point is that it operates on prepared tensors and captures the computation that     
 cannot be implemented efficiently by simply composing standard PyTorch operations.                                                                                                                                                     
                                                                                                                                                                                                                                        
 If, while writing the example, we discover that there is one canonical preparation operation that every user would otherwise have to reimplement, we can consider exposing it as another primitive. Interleaved block compression  might be such a candidate, but I would keep it private initially unless we can explain why it is independently useful and why its semantics are general enough to support future implementations. We should be very judicious here: a helper should not become public simply because it exists in the reference implementation.                                                                                                                                              
                           
                                                                                                                                                                                                                                        
 A concrete way to structure the work would be:                                                                                                                                                                                         
                                                                                                                                                                                                                                        
 1. Identify and document the minimal optimized primitive boundary. My current guess is the joint local-window and selected-compressed-block attention represented approximately by _launch_selected_attention.                         
                                                                                                                                                                                                                                        
 2. Give that primitive a clear eager/reference implementation whose inputs already satisfy the documented normalization, positional encoding, compression, and routing requirements.                                                   
                                                                                                                                                                                                                                        
 3. Adapt the Triton implementation and custom backward to implement that same primitive.                                                                                                                                               
                                                                                                                                                                                                                                        
 4. Keep preparation helpers private unless they are required for performance or have a clear, independently useful semantic contract.                                                                                                  
                                                                                                                                                                                                                                        
 5. Move the full compressed sparse attention recipe into a documented example that constructs the prepared inputs and calls the primitive.                                                                                             
                                                                                                                                                                                                                                        
 6. Test the primitive independently, including local-only attention, selected-block-only attention, their joint normalization, the attention sink, repeated selections, and forward/backward parity.                                   
                                                                                                                                                                                                                                        
 7. Add an end-to-end test for the example to demonstrate that the smaller API can still express the complete algorithm correctly.                                                                                                      
                                                                                                                                                                                                                                        
 The question I would like us to ask for every proposed public function and argument is: why must this be part of an Attention Gym API instead of ordinary PyTorch code that prepares inputs for the specialized operation? If the answer is only that the current reference implementation happens to contain it, that is not enough.                                                                                                                                    
                                                                                                                                                                                                                                        
 This approach would still give users an optimized and useful building block, while avoiding a large model-specific API that we will have to maintain and extend indefinitely. It also gives us room to learn from the next few compressed or sparse attention implementations before deciding whether there is a broader compressed-tensor abstraction worth exposing.

test/test_compressed_sparse_attention_triton.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is required")
+@pytest.mark.parametrize(("share_kv", "overrides"), PATHOLOGICAL_SHAPES)
+def test_triton_pathological_shapes_match_reference(share_kv, overrides):
+    configuration = {"batch": 1, **overrides}
+    assert_matches_reference(make_inputs(share_kv, torch.bfloat16, **configuration))
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is required")
```
drisspg: I have grown opinionated on testing over the years; for tldr is if we support fp32 and lower; we should compute using the reference in fp64; compute reference in kernel_dtype and then run kernel in kernel_dtype. Use the two references to establish a measuring stick and then compare your kernel to fp64 using that mearusing stick

attn_gym/sparse/compressed_sparse_attention/reference.py
floatingtrees: A lot of these functions are either mentioned in the paper or the huggingface inference code.

floatingtrees:
Hi Driss, I think that makes sense. I'll move the current reference into the examples folder and put the trimmed down version into a selected attention folder.

floatingtrees:
@drisspg I factored out the selected_attention part, implemented CSA in the examples folder, and wrote the tests you requested. Is there anything else I should add (more documentation to the CSA folder, broader test coverage, etc)?

attn_gym/sparse/compressed_sparse_attention/reference.py
floatingtrees: Ok, I have added doc blocks to all the reference functions.

test/test_compressed_sparse_attention_triton.py
floatingtrees: I added a test in `test/test_selected_attention_triton.py` that performs this comparison.

floatingtrees:
Hi Driss, here is a more detailed update:

I agree that the selected attention primitive the correct boundary, since most operations outside of it can be expressed fairly comfortably with standard PyTorch functions. The selected attention primitive does not perform normalization or RoPE. 

This PR contains a reference and triton implementation for selected attention. All functions in the reference have docstrings. Currently, only selected attention is public. I deleted most of the CSA tests and replaced them with tests for selected attention; these cover local-only attention, selected-block-only attention, their joint normalization, the attention sink, repeated selections, and forward/backward parity. There is also a test that compares lower precision error for the Triton kernel to lower precision error for the reference kernel. 

CSA has been moved into the `examples/` folder, updated to use the selected attention primitive and more documentation has been added for its helper functions. There's a test that imports CSA and compares it to the old reference. 

For the questions in the earlier comments, here is what I was thinking. 

> How is the query prepared? 

The query is projected out from the residual stream (ie `H @ W_q`), followed by optional RoPE and normalization. 

> How are the local and compressed KV tensors prepared?

Both computed as projections of the residual stream (with optional RoPE and normalization), followed by compression for compressed KV when applicable. Since we aren't compressing in the selected_attention function anymore, I refactored the code to rename compressed KV as index_KV, so the user can use compressed KV = local KV if they wish to. 

> How are the selected block indices produced? 

They should usually be produced with a topk function or something similar, though any discrete integer tensor with valid indices will work. 

> Which normalization and positional encoding conventions are required?

Selected attention does not perform any normalization or positional encoding; everything depends on the user. 

> Do any of these preparation steps require custom kernels for performance? 

At the moment, it seems like these steps do not require custom kernels for performance. 

Also, document masking support has been added to the reference and triton implementations.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
+from typing import Literal
+
+import torch
+
+
+Backend = Literal["eager", "triton", "cute"]
+Mode = Literal["auto", "chunked", "recurrent"]
```
drisspg: don need this

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def _validate_inputs(
+    Q: object,
+    KV: object,
```
drisspg: why are the types object here? that is werid

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+def selected_attention(
+    Q,
+    KV,
+    index_kv,
+    indices,
+    attention_sink,
+    doc_ids=None,
+    sliding_window_size: int = 512,
+    share_kv: bool = True,
+    backend: Backend = "eager",
```
drisspg: I think the default backend should be `"triton"` rather than `"eager"`.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+
+def selected_attention(
+    Q,
+    KV,
+    index_kv,
+    indices,
+    attention_sink,
+    doc_ids=None,
+    sliding_window_size: int = 512,
+    share_kv: bool = True,
```
drisspg: Could we remove `share_kv` and infer this from the number of heads in `Q`, `KV`, and `index_kv`? This information is already encoded in the tensor shapes, so I don't think users should need to specify it separately unless there is an implementation constraint that makes inference harder.

drisspg (commented):
Some comments

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+            represented as a shared tensor, (batch_size, 1, sequence_length, head_dim) if share_kv = False
+            Otherwise represented as (batch_size, num_heads, sequence_length, head_dim)
+
+        index_kv: KV for the indexing branch, shape of (batch, 1, X, head_dim) if share_kv = False
+            Otherwise represented as (batch_size, num_heads, X, head_dim)
+            where X is any number greater than
+
+        indices: Which indices to attend to. Shape of (batch, sequence_length, num_topk_blocks), integer tensor
+            If None, index_kv will be ignored
+
```
drisspg, line 170: I don't think `indices=None` needs to be part of this API. If there is no `index_kv`, this is just sliding-window attention, which users can already do through the existing FlexAttention or varlen paths and do not need this primitive for. My understanding is that the point of `selected_attention` is specifically to combine local-window and indexed KV under the same normalization, so I think `index_kv` and `indices` should be required.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+    KV,
+    index_kv,
+    indices,
+    attention_sink,
+    doc_ids=None,
+    sliding_window_size: int = 512,
+    share_kv: bool = True,
+    backend: Backend = "eager",
+    mode: Mode = "auto",
+):
```
drisspg: Could we start this docstring by defining the operation semantically? The key contract, as I understand it, is that each query attends jointly to its causal local window and the entries selected from `index_kv`, with both branches and the attention sink participating in one shared softmax normalization. That distinction is important because computing the two attention branches independently would produce different results.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+        or KV.shape[1] != expected_kv_heads
+        or KV.shape[2] != sequence_length
+        or KV.shape[3] != head_dim
+    ):
+        expected_h = "1 or heads" if share_kv else "heads"
+        raise ValueError(
+            f"KV must have shape [batch, {expected_h}, sequence_length, head_dim], "
+            f"got {list(KV.shape)}."
+        )
+    index_seq_len = index_kv.shape[2]
```
drisspg: Could we validate `index_kv` here in the same way as `KV`? We should be able to check its rank, shape, and dtype.

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,431 @@
...
+        index_kv_lp_tri,
+        indices,
+        sink_lp_tri,
+        None,
+        sliding_window_size,
+        share_kv,
+        backend="triton",
+    )
+
+    ref_fwd_diff = (out_64.float() - out_lp_ref.float()).abs().max().item()
```
drisspg: Could we compute these errors in fp64 rather than casting the fp64 reference to fp32? This should follow the FlexAttention testing pattern: keep the golden result in fp64 and promote the lower-precision eager and Triton results to fp64 before computing their errors. The same applies to the gradient comparisons.

test/test_selected_attention_semantics.py
```diff
@@ -0,0 +1,422 @@
...
+    scores = torch.randn(b, s, 6, device=device)
+    _, indices = torch.topk(scores, k=2, dim=-1)
+    out_joint = selected_attention(
+        Q, KV, index_kv, indices, sink, None, window, False, backend=backend
+    )
+
+    # They should differ (joint normalization changes the local contribution)
+    assert not torch.allclose(
+        out_local, out_joint, atol=1e-3
+    ), "Adding index blocks should change output due to shared normalization"
```
drisspg: I don’t think this test establishes that the two branches share one normalization. Could we compare against an independent manual computation that concatenates the local and selected logits and applies a single softmax, including the sink? Otherwise, an implementation using separate per-branch softmax operations could still pass this test.

test/test_selected_attention_semantics.py
```diff
@@ -0,0 +1,422 @@
...
+    sink = torch.randn(h, device=device, dtype=dtype)
+
+    # All slots repeat position 2
+    indices = torch.full((b, s, num_repeats), 2, dtype=torch.long, device=device)
+
+    out_eager = selected_attention(
+        Q, KV, index_kv, indices, sink, None, sliding_window_size, False, backend="eager"
+    )
+    out_triton = selected_attention(
+        Q, KV, index_kv, indices, sink, None, sliding_window_size, False, backend="triton"
```
drisspg: Could we test repeated-index semantics against an independent manual result rather than only comparing the eager and Triton implementations? Otherwise, both implementations could handle duplicates incorrectly in the same way. I think we should also include a backward check here, since repeated selections are particularly relevant to gradient accumulation into `index_kv`.

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,431 @@
...
+        f"\n  {f'ref {dtype_name}':15s} {ref_fwd_diff:10.4e} {ref_dq:10.4e} {ref_dkv:10.4e}"
+        f" {ref_didx:10.4e} {ref_dsink:10.4e}"
+        f"\n  {f'triton {dtype_name}':15s} {tri_fwd_diff:10.4e} {tri_dq:10.4e} {tri_dkv:10.4e}"
+        f" {tri_didx:10.4e} {tri_dsink:10.4e}"
+        f"\n  {'triton/ref':15s} {r_fwd:10.2f}x {r_dq:10.2f}x {r_dkv:10.2f}x"
+        f" {r_didx:10.2f}x {r_dsink:10.2f}x"
+    )
+
+    # Sanity checks to make sure we don't get anything insane
+    # Not really meant for checking correctness
```
drisspg: How did we arrive at `5` here? Is this as low as the threshold can reasonably be?

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+        Tensor in shape of (batch_size, num_heads, sequence_length, head_dim)
+    """
+    if not torch.compiler.is_compiling():
+        _validate_inputs(
+            Q, KV, index_kv, indices, attention_sink, doc_ids, sliding_window_size, share_kv
+        )
+
+    match backend:
+        case "eager":
+            from . import reference
```
drisspg: Could we move the eager and Triton implementations under `selected_attention/impl/` as described in `DESIGN.md`? I would like this first implementation to establish the intended separation between the public API and private backend implementations.

test/test_selected_attention_matches_csa_reference.py
```diff
@@ -0,0 +1,391 @@
...
+) -> torch.Tensor:
+    sequence_length = x.shape[-2]
+    rotary_dim = x.shape[-1]
+
+    if positions is None:
+        positions = torch.arange(
+            position_offset,
+            position_offset + sequence_length,
+            device=x.device,
+            dtype=torch.float32,
```
drisspg: Is there a reason the fp64 reference performs RoPE in fp32 here? Since this test is intended to provide an fp64 end-to-end reference, I would expect the RoPE computation to preserve the input dtype unless the fp32 behavior is part of the algorithm’s required semantics.

attn_gym/sparse/selected_attention/reference.py
```diff
@@ -0,0 +1,169 @@
+"""Torch-only compressed sparse attention reference implementation."""
+
+import torch
+
+
+def make_sliding_window_mask(query_length, window_size, device, dtype):
```
drisspg: Could we add complete type annotations here and use the docstring only for semantics, shapes, and constraints? For example, `query_length: int`, `window_size: int`, `device: torch.device`, `dtype: torch.dtype`, and `-> torch.Tensor`. The docstring can then describe the causal sliding-window convention and returned mask shape without repeating those types.

attn_gym/sparse/selected_attention/impl/reference.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    query_positions = torch.arange(query_length, device=device)[:, None]
+    key_positions = torch.arange(query_length, device=device)[None, :]
+    valid = (key_positions <= query_positions) & (
+        key_positions >= query_positions - window_size + 1
+    )
+    return torch.zeros(
+        (query_length, query_length),
+        device=device,
+        dtype=dtype,
```
drisspg, line 28: Could we type this signature and make the docstring precise about the sink semantics? In particular, it would be useful to say that the sink contributes an additional logit to the softmax denominator but has no corresponding value and therefore is not present in the returned tensor.

attn_gym/sparse/selected_attention/impl/reference.py
```diff
@@ -0,0 +1,169 @@
...
+        device=device,
+    )
+    mask.masked_fill_(same_document, 0.0)
+
+    # Head-broadcasting dimension.
+    return mask[:, None, :, :]
+
+
+def selected_attention(
```
drisspg, line 66: Could we add complete type annotations to this reference entry point and rewrite this as a Google-style semantic docstring without repeating types from the signature? The docstring should precisely define the joint local/indexed softmax, tensor shapes, `-1` sentinel and repeated-index behavior, document masking responsibility, and the returned shape. There are also incomplete/reversed descriptions here, such as “where X is any number greater than” and the `share_kv` shape descriptions.

attn_gym/sparse/selected_attention/impl/reference.py
```diff
@@ -0,0 +1,169 @@
...
+        sink: tensor in shape of (num_heads, )
+        dim: integer, dimension to apply softmax on
+    Returns:
+        tensor in shape of x, where tensor[a, b, c, d] = exp(x[a, b, c, d]) / (sum(exp(x[a, b, c, :]) + exp(sink[b])))
+    """
+    sink = sink[None, :, None, None]
+    maximums = torch.max(x, dim=dim, keepdim=True).values
+    maximums = torch.maximum(maximums, sink)
+    x = x - maximums
+    sink = sink - maximums
```
drisspg: Could we add a one-line summary before `Args:` and document the returned mask semantics more precisely? With the types in the signature, the docstring only needs to explain that equal document IDs are allowed and different document IDs receive `-inf`, plus the `[batch, 1, query, key]` output shape.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+    if type(share_kv) is not bool:
+        raise TypeError(f"share_kv must be a Python bool, got {type(share_kv).__name__}.")
+
+    tensors = {
+        "Q": Q,
+        "KV": KV,
+        "index_kv": index_kv,
+        "indices": indices,
+        "attention_sink": attention_sink,
```
drisspg, line 25: Could we also validate that the floating-point inputs have the same dtype and device as `Q`?

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,431 @@
...
+    inputs = _make_inputs(share_kv=share_kv, num_topk=num_topk, seq_len=seq_len, doc_ids=doc_ids)
+
+    with torch.inference_mode():
+        expected = selected_attention(**inputs, backend="eager")
+        actual = selected_attention(**inputs, backend="triton")
+
+    torch.testing.assert_close(actual, expected, atol=ATOL_FWD, rtol=RTOL_FWD)
+
+
+@pytest.mark.parametrize("share_kv", [False, True])
```
drisspg: Could we add a backward parity case with `sliding_window_size=0` and nonempty `indices`? We currently cover selected-only forward, but not selected-only backward.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+            raise TypeError(
+                f"doc_ids must be a torch.Tensor or None, got {type(doc_ids).__name__}."
+            )
+        if doc_ids.ndim != 2 or doc_ids.shape[0] != batch or doc_ids.shape[1] != sequence_length:
+            raise ValueError(
+                f"doc_ids must have shape [batch, sequence_length], " f"got {list(doc_ids.shape)}."
+            )
+
+
+def selected_attention(
```
drisspg, line 139: Could we use clearer names for the public API?

```python
selected_attention(
    query,
    local_kv,
    sparse_kv,
    kv_indices,
    attention_sink,
    ...
)
```

`local_kv` identifies the sliding-window branch, `sparse_kv` identifies the candidate KV pool, and `kv_indices` selects entries from that pool. I find this clearer than `Q`, `KV`, `index_kv`, and `indices`.

attn_gym/sparse/selected_attention/api.py
```diff
@@ -0,0 +1,184 @@
...
+    Q,
```
drisspg: Could we fully type the public signature, including the tensor arguments, optional `doc_ids`, and return type? We can use `from torch import Tensor` so the annotations stay concise and the docstring can focus on semantics and shapes.

attn_gym/sparse/selected_attention/triton.py
```diff
@@ -0,0 +1,979 @@
...
+    """
+    b, h, s, head_dim = Q.shape
+
+    # Validate CUDA
+    if Q.device.type != "cuda":
+        raise ValueError("The Triton selected attention backend requires CUDA tensors.")
+
+    # Expand shared KV heads
+    if share_kv:
+        KV = KV.expand(-1, h, -1, -1).contiguous()
```
drisspg: Why do we need to materialize the shared KV tensors across all query heads here? Could the kernel map each query head to the shared KV head directly, or support a zero head stride from `expand`, rather than calling `contiguous()` and duplicating the data? I’m concerned this could create significant memory overhead for realistic head counts.

attn_gym/sparse/selected_attention/impl/reference.py
drisspg, line 28: Could we simplify the reference by concatenating the sink logit, calling `torch.softmax`, and dropping the sink column? That seems like a more direct expression of the semantics than maintaining a custom softmax implementation.

attn_gym/sparse/selected_attention/impl/triton.py
```diff
@@ -0,0 +1,979 @@
...
+        grad_values,
+        mask=key_mask[:, None] & dimension_mask[None, :],
+    )
+
+
+@triton.jit
+def _selected_attention_bwd_dindex_kv(
+    q_ptr,
+    index_kv_ptr,
+    indices_ptr,
```
drisspg: It looks like this kernel never reads `indices_ptr`, its strides, or `INDEX_SEQ_LEN`. Could we remove those arguments, or add a comment explaining why they still need to be passed?

attn_gym/sparse/selected_attention/impl/triton.py
```diff
@@ -0,0 +1,979 @@
...
+    block_n = 128
+    block_d = max(16, triton.next_power_of_2(head_dim))
+    num_local_tiles = (
+        triton.cdiv(sliding_window_size + block_m - 1, block_n) if sliding_window_size else 0
+    )
+
+    output = torch.empty_like(Q)
+    lse = torch.empty(batch, heads, seq_len, device=Q.device, dtype=torch.float32)
+
+    has_doc_ids = doc_ids is not None
```
drisspg, line 655: Could we avoid allocating a dummy CUDA tensor when `doc_ids` is absent? Since `HAS_DOC_IDS=False` removes the loads, it seems like we could pass an existing tensor as the unused pointer with zero strides. The backward launcher has the same allocation.

test/test_selected_attention_semantics.py
```diff
@@ -0,0 +1,422 @@
...
+    KV_c = KV_ref.detach().clone().requires_grad_(True)
+    idx_c = idx_ref.detach().clone().requires_grad_(True)
+    sink_c = sink_ref.detach().clone().requires_grad_(True)
+
+    out_c = compiled_fn(Q_c, KV_c, idx_c, sink_c)
+    out_c.sum().backward()
+
+    torch.testing.assert_close(out_c, out_ref, atol=1e-5, rtol=1e-5)
+    torch.testing.assert_close(Q_c.grad, Q_ref.grad, atol=1e-5, rtol=1e-5)
+    torch.testing.assert_close(KV_c.grad, KV_ref.grad, atol=1e-5, rtol=1e-5)
```
drisspg: Since `idx_ref`/`idx_c` and `sink_ref`/`sink_c` require gradients, could we check their gradients here as well? Otherwise, the compile-backward test does not cover all differentiable inputs it constructs.

test/test_selected_attention_reference.py
```diff
@@ -0,0 +1,277 @@
...
+        Q2, KV2, idx_kv2, indices2, attention_sink, None, sliding_window_size, share_kv,
+    )
+
+    # ---------- Pack into one sequence ----------
+    Q_packed = torch.cat([Q1, Q2], dim=2)
+    KV_packed = torch.cat([KV1, KV2], dim=2)
+
+    # For the index branch, pack both index KVs and adjust indices for doc2
+    idx_kv_packed = torch.cat([idx_kv1, idx_kv2], dim=2)
+    packed_idx_len = idx_kv1_len + idx_kv2_len
```
drisspg: `packed_idx_len`, `indices1_padded`, and `indices2_padded` appear to be unused scaffolding. Could we remove them and construct only `indices_packed`?

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,427 @@
+"""
```
drisspg: This docstring references `test_reference_bf16_vs_fp64_precision`, but the test is named `test_precision_vs_fp64`. Could we remove this diagnostic header or update it so it does not become stale?

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,427 @@
...
+    inputs_tri = _make_inputs(share_kv=share_kv, num_topk=num_topk, requires_grad=True, seed=456)
+
+    out_ref = selected_attention(**inputs_ref, backend="eager")
+    out_tri = selected_attention(**inputs_tri, backend="triton")
+
+    grad_gen = torch.Generator(device=out_ref.device).manual_seed(8888)
+    grad_output = torch.randn(out_ref.shape, device=out_ref.device, generator=grad_gen)
+    out_ref.backward(grad_output)
+    out_tri.backward(grad_output)
+
```
drisspg: These backward tests repeat the same input construction, forward/backward execution, and gradient comparison. Could we parameterize the gradient target and relevant cases, or extract a small test helper, so the coverage remains easier to extend and maintain?

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, I removed it

attn_gym/sparse/selected_attention/api.py
floatingtrees: I think they were relics of placeholder code, and I replaced them with proper types now.

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, I changed it.

attn_gym/sparse/selected_attention/api.py
floatingtrees: That makes sense, I removed it from the API

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, I added a description to the operation.

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, I added checks for it.

test/test_selected_attention_triton.py
floatingtrees: That makes sense, everything is cast to fp64 now.

test/test_selected_attention_matches_csa_reference.py
floatingtrees: Thanks for catching this, I fixed it.

attn_gym/sparse/selected_attention/reference.py
floatingtrees: Ok, I fixed it.

attn_gym/sparse/selected_attention/impl/reference.py
floatingtrees: Ok, I updated the docstring.

floatingtrees:
Hi Driss, thanks for the detailed comments, I've started fixing them. I'll reply to individual comments as they're fixed or if I have questions.

attn_gym/sparse/selected_attention/impl/reference.py
floatingtrees: Ok, I fixed it.

attn_gym/sparse/selected_attention/impl/reference.py
floatingtrees: That makes sense, I updated it.

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, fixed

test/test_selected_attention_triton.py
floatingtrees: There's a similar test case in `test_selected_attention_semantics`; I moved it here.

attn_gym/sparse/selected_attention/api.py
floatingtrees: These names seem clearer to me as well; I updated the names.

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, I typed it.

test/test_selected_attention_semantics.py
floatingtrees: That makes sense, I added a manual test with small shapes and a backwards test for triton.

test/test_selected_attention_triton.py
floatingtrees: In practice, triton seems to be between 0.59-1.23x the reference bf16 diff, so we might be able to reduce it to 2x. The 5 is somewhat arbitrary; do you have any suggestions? I saw that in flex attention this value ranged from  1.1 to 10.

attn_gym/sparse/selected_attention/api.py
floatingtrees: Ok, fixed.

attn_gym/sparse/selected_attention/api.py
floatingtrees: That makes sense, I fixed it.

attn_gym/sparse/selected_attention/triton.py
floatingtrees: That makes sense, I fixed it for the forward pass. In the backwards pass, we need to perform a reduction to compute dKV, so I think materializing a large tensor might be preferable to using atomics. What do you think?

attn_gym/sparse/selected_attention/impl/reference.py
floatingtrees: Ok, I fixed it.

test/test_selected_attention_triton.py
floatingtrees: That makes sense, I parameterized these test cases.

test/test_selected_attention_triton.py
floatingtrees: Thanks for the catch, I fixed it.

test/test_selected_attention_reference.py
floatingtrees: I removed them.

test/test_selected_attention_semantics.py
floatingtrees: Ok, I added a check

attn_gym/sparse/selected_attention/impl/triton.py
floatingtrees: That makes sense, I'm passing in a zero stride tensor now.

attn_gym/sparse/selected_attention/impl/triton.py
floatingtrees: Thanks for the catch, I removed them.

test/test_selected_attention_semantics.py
floatingtrees: Thanks for the catch, I updated the function

floatingtrees:
@drisspg I've fixed all of the comments with two partial exceptions: for the contiguous() calls in the triton kernel, I removed it from forward, since it's read only, but for backward(), we need to perform a reduction to compute dKV, so I kept it, since the alternative would be to use atomics. Also, I left the fp32 to bf64 ratio in the test as 5, which is a somewhat arbitrary value, since I didn't know if we could meaningfully replace it with a less arbitrary value. Is there anything else that should be fixed?

attn_gym/sparse/selected_attention/triton.py
drisspg: yeah I think thats fine we do a similar thing for our flex triton impl

drisspg:
Sounds good thanks for all the fixes

drisspg (approved):

drisspg:
can you run prek and get things formatted then ready to merge

floatingtrees:
@drisspg `prek run --all-files` works and all relevant tests pass, so it should be good to merge now.

PR #239, opened 2026-08-06
Added a benchmarking script for selected attention

floatingtrees:
Separate PR from the benchmarking script, as requested by #225. The benchmark script is under `benchmarks/sparse/selected_attention_benchmark.py`. It exists to measure the relative performance of different selected attention backends under expected usage conditions. Currently, the numbers are as follows for the hyperparameters in the script:
Trition: 
```
forward: 3.538 ms  (80.12 TFLOP/s)
```
Eager: 
```
forward: 66.348 ms  (4.27 TFLOP/s)
```

### DIFF
--- a/benchmarks/sparse/selected_attention_benchmark.py
+++ b/benchmarks/sparse/selected_attention_benchmark.py
@@ -1,0 +1,177 @@
+"""Benchmark selected_attention forward and backward across backends and shapes.
+
+Usage:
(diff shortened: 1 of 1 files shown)


drisspg (approved):
can you check after the latest 2 prs I just landed they should have helped perf at least on b200

floatingtrees:
They helped a lot, performance is now: 
```
forward: 1.047 ms  (270.86 TFLOP/s)
```

PR #241, opened 2026-08-07
add tests for kda kernels + naive impls for gate/l2norm/cumsum

liangel-02:
as title, all tests pass, using the golden ref trick (instead of hardcoding a tolerance) we have for varlen/flex to assert, also fixing lint from previous pr

### DIFF
--- a/attn_gym/linear/kda/naive.py
+++ b/attn_gym/linear/kda/naive.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from itertools import pairwise
+
 import torch
 import torch.nn.functional as F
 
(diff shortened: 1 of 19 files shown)


attn_gym/linear/kda/naive.py
```diff
@@ -124,4 +126,117 @@ def to_chunks(x):
     return o, (state.to(orig_dtype) if output_final_state else None)
 
 
-__all__ = ["naive_chunk_kda", "naive_recurrent_kda"]
+def l2norm_fwd_ref(
+    x: torch.Tensor,
+    eps: float = 1e-6,
+    compute_dtype: torch.dtype = torch.float32,
```
drisspg: hmm do we need compute_dtype ; liek coudl it jsut bed defineid intenrally as max(torch.float32, x.dtype) or equiavelnt if x.dtype is fp64 since literally I dont think max would work

attn_gym/linear/kda/naive.py
```diff
@@ -124,4 +126,117 @@ def to_chunks(x):
...
+
+
+def chunk_cumsum_ref(
+    x: torch.Tensor,
+    chunk_size: int,
+    reverse: bool = False,
+    scale: float | None = None,
+    cu_seqlens: torch.Tensor | None = None,
+) -> torch.Tensor:
+    """Chunk-local cumsum over the time axis (dim=1), optionally per document.
```
drisspg, line 171: maybe note on the expected shape of X e.g. (B, N, dim) or whatever it is

attn_gym/linear/kda/naive.py
```diff
@@ -124,4 +126,117 @@ def to_chunks(x):
...
+    reverse: bool,
+    chunk_size: int,
+    cu_seqlens: torch.Tensor | None,
+    compute_dtype: torch.dtype,
+) -> torch.Tensor:
+    """Gate map, then chunk-local cumsum.
+
+    Naive counterpart of ``kda_gate_chunk_cumsum_vector_kernel`` (fused gate + chunk cumsum).
+    """
+    z = g.to(compute_dtype)
```
drisspg: actually mabye for all these dtypes; I think all teh publci funcs shoudl desribe the semantics too of the args

test/test_kda_kernels.py
```diff
@@ -0,0 +1,343 @@
...
+    bias64 = torch.randn(H, S, device=DEV, dtype=torch.float64) if has_bias else None
+
+    golden = gate_fwd_ref(
+        g64, A_log64, bias64, lower_bound, scale, False, chunk, None, torch.float64
+    )
+
+    g = g64.to(dtype)
+    A_log = A_log64.to(torch.float32)
+    bias = bias64.to(dtype) if has_bias else None
+    ref = gate_fwd_ref(g, A_log, bias, lower_bound, scale, False, chunk, None, dtype)
```
drisspg: The low-precision gate reference uses `compute_dtype=dtype`, but both Triton kernels explicitly compute in FP32. That introduces extra FP16/BF16 rounding—especially by downcasting the FP32 `A_log`—and inflates the golden-error budget. Could we pass `compute_dtype=torch.float32` here and in the backward reference so the numerics match the kernel’s actual inputs and accumulation semantics?

test/test_kda_kernels.py
```diff
@@ -0,0 +1,343 @@
...
+    rstd = torch.empty(T, device=DEV, dtype=torch.float32)
+    grid = lambda meta: (triton.cdiv(T, meta["BT"]),)
+    l2norm_fwd_kernel[grid](x, y, rstd, eps, T, D=D, BD=BD, NB=triton.cdiv(T, 16))
+    assert_golden(y, golden, ref, dtype, f"l2norm_fwd_kernel T={T} D={D}")
+
+    # single-row kernel
+    y1 = torch.empty_like(x)
+    rstd1 = torch.empty(T, device=DEV, dtype=torch.float32)
+    l2norm_fwd_kernel1[(T,)](x, y1, rstd1, eps, D=D, BD=BD)
+    assert_golden(y1, golden, ref, dtype, f"l2norm_fwd_kernel1 T={T} D={D}")
```
drisspg, line 112: Neither forward L2-normalization test checks `rstd`/`rstd1`. These saved kernel outputs are consumed by backward, so incorrect statistics could go unnoticed while `y` remains within tolerance. Please compare both against the FP32 reference `torch.rsqrt(torch.sum(x.float() ** 2, dim=-1, keepdim=True) + eps)`.

test/test_kda_kernels.py
```diff
@@ -0,0 +1,343 @@
...
+        T,
+        None,
+        B=B,
+        H=H,
+        S=S,
+        S_in=S,
+        F_REPEAT=1,
+        BT=chunk_size,
+        REVERSE=False,
+    )
```
drisspg: The gate-forward launcher always uses `REVERSE=False`, `S_in=S`, and `F_REPEAT=1`, so the reverse-cumsum and repeated-input branches remain untested. At minimum, please parameterize `REVERSE`; ideally, also add a case with `S_in < S`, since these branches use distinct indexing logic.

drisspg (approved):
Some nits and ai found edge cases

PR #242, opened 2026-08-07
Fixed accidental deletion of triton test due to linting

floatingtrees:
Somehow the test cases for triton were delete on the last commit of #232. This PR brings them back. All tests in `test/test_selected_attention_triton.py` pass, and `prek run -a` passes as well.

### DIFF
--- a/test/test_selected_attention_triton.py
+++ b/test/test_selected_attention_triton.py
@@ -1,0 +1,403 @@
+"""
+To check reference bf16 vs triton bf16 diffs, run:
+python -m pytest test/test_selected_attention_triton.py::test_precision_vs_fp64 -v -s
+
+Triton max diff is between 0.59-1.23x the reference's max diff
+"""
+
+import pytest
+import torch
+
+from attn_gym.sparse.selected_attention import selected_attention
+
+ATOL_FWD = 1e-2
+RTOL_FWD = 1e-2
+ATOL_BWD = 1e-2
+RTOL_BWD = 1e-2
+
+
+def _skip_no_cuda():
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA required for triton backend")
+
+
+def _make_inputs(
+    batch: int = 2,
+    heads: int = 4,
+    seq_len: int = 32,
+    head_dim: int = 64,
+    sparse_seq_len: int = 16,
+    num_topk: int = 3,
+    sliding_window_size: int = 8,
+    share_kv: bool = True,
+    doc_ids: torch.Tensor | None = None,
+    dtype: torch.dtype = torch.float32,
+    requires_grad: bool = False,
+    seed: int = 42,
+):
+    device = torch.device("cuda")
+    generator = torch.Generator(device=device).manual_seed(seed)
+    kv_heads = 1 if share_kv else heads
+
+    def randn(*shape):
+        return torch.randn(
+            *shape, device=device, dtype=dtype, generator=generator, requires_grad=requires_grad
+        )
+
+    query = randn(batch, heads, seq_len, head_dim)
+    local_kv = randn(batch, kv_heads, seq_len, head_dim)
+    sparse_kv = randn(batch, kv_heads, sparse_seq_len, head_dim)
+
+    if num_topk > 0:
+        scores = torch.randn(batch, seq_len, sparse_seq_len, device=device, generator=generator)
+        _, kv_indices = torch.topk(scores, k=min(num_topk, sparse_seq_len), dim=-1)
+    else:
+        kv_indices = torch.zeros(batch, seq_len, 0, dtype=torch.long, device=device)
+
+    attention_sink = torch.randn(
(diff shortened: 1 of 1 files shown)


test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,403 @@
...
+    _skip_no_cuda()
+    batch, heads, seq_len, head_dim = 2, 4, 32, 64
+    sparse_seq_len = 16
+    sliding_window_size = 8
+    kv_heads = 1 if share_kv else heads
+    seed = 77
+    device = torch.device("cuda")
+    dtype_name = str(dtype).removeprefix("torch.")
+
+    # --- Generate inputs in fp64 (ground truth) ---
```
drisspg: Could we flip how these inputs are constructed? The measuring-stick pattern should create the low-precision inputs once, then derive the FP64 inputs with `.double()`, so eager, Triton, and FP64 all see the same quantized values. Starting in FP64 and casting down makes the baseline include input-quantization error, which can hide kernel error. There is already an example of the intended pattern in `test_selected_attention_semantics.py::assert_matches_low_precision_eager` and its callers.

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,403 @@
...
+        sliding_window_size,
+        backend="triton",
+    )
+
+    ref_fwd_diff = (out_64 - out_lp_ref).abs().max().item()
+    tri_fwd_diff = (out_64 - out_lp_tri).abs().max().item()
+
+    # --- Backward ---
+    grad_gen = torch.Generator(device=device).manual_seed(1234)
+    grad_64 = torch.randn(out_64.shape, dtype=torch.float64, device=device, generator=grad_gen)
```
drisspg: Same issue for backward: could we generate the upstream gradient in `dtype` first and use `grad_lp.double()` for the FP64 reference? Otherwise the reported gradient error also includes quantization of the cotangent, rather than isolating the eager-vs-kernel numerical error on identical inputs.

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,403 @@
...
+        f" {ref_didx:10.4e} {ref_dsink:10.4e}"
+        f"\n  {f'triton {dtype_name}':15s} {tri_fwd_diff:10.4e} {tri_dq:10.4e} {tri_dkv:10.4e}"
+        f" {tri_didx:10.4e} {tri_dsink:10.4e}"
+        f"\n  {'triton/ref':15s} {r_fwd:10.2f}x {r_dq:10.2f}x {r_dkv:10.2f}x"
+        f" {r_didx:10.2f}x {r_dsink:10.2f}x"
+    )
+
+    # Sanity checks to make sure we don't get anything insane
+    # Not really meant for checking correctness
+    assert r_fwd < 5, (
```
drisspg: I don't think an arbitrary `5x` ratio is a useful correctness gate; this explicitly allows the Triton result to be almost five times worse than low-precision eager. Could we use the existing measuring-stick style instead: compare both paths to FP64, require Triton error to be no worse than eager plus a justified accumulation/output-rounding allowance, and check finiteness plus both mean and max/pointwise error?

test/test_selected_attention_triton.py
```diff
@@ -0,0 +1,403 @@
...
+        query, local_kv, sparse_kv, kv_indices, sink, None, sliding_window_size, backend="eager"
+    )
+    out_triton = selected_attention(
+        query, local_kv, sparse_kv, kv_indices, sink, None, sliding_window_size, backend="triton"
+    )
+
+    torch.testing.assert_close(out_eager, out_triton, atol=1e-4, rtol=1e-4)
+
+
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
```
drisspg, line 253: Could we include FP16 in the FP64 measuring-stick test below (including gradients), rather than only checking it directly against low-precision eager with a broad `5e-2` tolerance? As written, FP16 does not get the stronger numerical baseline at all.

test/test_selected_attention_triton.py
floatingtrees: That makes sense, I fixed it.

test/test_selected_attention_triton.py
floatingtrees: Fixed as well.

test/test_selected_attention_triton.py
floatingtrees: I switched to using the current method in `test_selected_attention_semantics.py::assert_matches_low_precision_eager`, but I added additional tolerance for max error to compensate for intermediate bf16 rounding (added bf16.eps to tolerance for each reduction performed, for max error tolerance only).

test/test_selected_attention_triton.py
floatingtrees: Ok, I added it.

floatingtrees:
Hi Driss, thanks for the feedback. I fixed the issues you mentioned. All tests pass, and prek passes too.

PR #261, opened 2026-08-10
kda tests

liangel-02:


test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    """
+
+    def tf32(x: torch.Tensor) -> torch.Tensor:
+        if x.dtype is not torch.float32:
+            return x
+        i = x.contiguous().view(torch.int32)
+        i = (i + 0x1000) & ~0x1FFF
+        return i.view(torch.float32)
+
+    return torch.einsum(eqn, tf32(a), tf32(b))
```
drisspg: why not jsut mm withthe right mode?

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    torch.manual_seed(10)
+    B, V = 1, 64
+    scale = 0.5
+    v64 = torch.randn(B, T, H, V, device="cuda", dtype=torch.float64)
+    do64 = torch.randn(B, T, H, V, device="cuda", dtype=torch.float64)
+    A64 = torch.randn(B, T, H, 64, device="cuda", dtype=torch.float64)
+
+    gdv, gdA = _bwd_dav_ref(v64, A64, do64, scale)
+
+    v, do, A = v64.to(dtype), do64.to(dtype), A64.to(dtype)
```
drisspg: yeah the agreed flow goes low to up 
Agent: 
Please create the low-precision inputs first, then derive the FP64 measuring-stick inputs with `.double()`. Starting from FP64 and casting down means the reference error includes input quantization, which inflates the kernel's allowed error. This pattern appears throughout the new Triton tests; all paths should operate on identical input values.

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    scale = K**-0.5
+    q64 = torch.randn(B, T, H, K, device="cuda", dtype=torch.float64)
+    # cumulative decay stays small so 2^g is well-scaled
+    g64 = torch.randn(B, T, H, K, device="cuda", dtype=torch.float64) * 0.1
+    h64 = torch.randn(B * num_chunks, H, K, V, device="cuda", dtype=torch.float64)
+    A64 = torch.randn(B, T, H, 64, device="cuda", dtype=torch.float64)
+    v64 = torch.randn(B, T, H, V, device="cuda", dtype=torch.float64)
+
+    golden = _fwd_o_ref(q64, g64, h64, A64, v64, scale)
+    q, g, h, A, v = (x.to(dtype) for x in (q64, g64, h64, A64, v64))
```
drisspg: seems legit

Could we keep `g` in FP32 here (and in the TMA case below)? The production path explicitly converts `cumulative_gate` to FP32 before calling this launcher, and the kernel applies `exp2` to those loaded values. Downcasting it with the other operands means these tests exercise a different ABI and can hide gate-precision issues.

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    # State recurrence is sequential across chunks; batch (B, H) into the matmuls.
+    state = torch.zeros(B, H, K, V, dtype=acc, device=k.device) if h0 is None else h0.to(acc)
+    for it in range(num_chunks):
+        s = it * chunk_size
+        e = min(s + chunk_size, T)
+        h[:, it] = state
+        vn = vc[:, s:e] - _mm("blhk,bhkv->blhv", wc[:, s:e], state)  # (B, L, H, V)
+        v_new[:, s:e] = vn
+        state = state * gkc[:, e - 1].exp2()[..., None]
+        state = state + _mm("blhk,blhv->bhkv", kc[:, s:e], vn)
```
drisspg: agreed here; tangentially I wanatto ahve a really cler description of the full data dtype flow for outpus and grads. Lets try and mimic where possible

I think the reference needs to model the kernel's quantization boundary here. The kernel casts the corrected `b_v` back to the input dtype before the `k.T @ v_new` state update, while this keeps `vn` in FP32. Could we cast `vn` to `k.dtype` for this update so the recurrent reference follows the actual computation?

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    # gk is the fused gate/cumsum output and is always fp32 (the kernel calls exp2 on it
+    # directly, without upcasting).
+    k, w, v = k64.to(dtype), w64.to(dtype), v64.to(dtype)
+    gk = gk64.to(torch.float32)
+    h0 = h0_64.to(dtype) if use_h0 else None
+    rh, rvn, rht = _fwd_h_ref(k, w, v, gk, h0)
+
+    h = torch.empty(B * triton.cdiv(T, 64), H, K, V, device="cuda", dtype=dtype)
+    v_new = torch.empty(B, T, H, V, device="cuda", dtype=dtype)
+    ht = torch.empty(B, H, K, V, device="cuda", dtype=dtype)
```
drisspg: Could we exercise the production state dtype here? `chunk_gated_delta_rule_fwd_h` allocates `final_state` as FP32, and the public KDA path normalizes a provided initial state to FP32. Using low-precision `h0`/`ht` tests a different load/store and recurrence contract, so FP32-state regressions can pass.

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+        chunk_indices=None,
+        num_chunks=None,
+        T=T,
+        H=H,
+        K=K,
+        BT=64,
+        BC=BC,
+        BK=triton.next_power_of_2(K),
+        USE_GATHER=False,
+        CAUSAL_NORMREF=True,
```
drisspg: This is not the specialization used by the production launcher: it passes `CAUSAL_NORMREF=False` and `USE_GATHER=IS_GATHER_SUPPORTED`. Could we test those production flags (including the gather path when available) rather than only the default/reference specialization?

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+        chunk_kda_bwd_wy_dqkg as chunk_kda_bwd_wy_dqkg_fused_cute,
+    )
+    from attn_gym.linear.kda.fwd.cute.chunk_kda_fwd_intra import (
+        chunk_kda_fwd_intra as chunk_kda_fwd_intra_cute,
+    )
+    from attn_gym.linear.kda.fwd.cute.recompute_w_u_fwd import recompute_w_u_fwd
+
+    HAS_CUTE = True
+    CUTE_IMPORT_ERR = ""
+except Exception as e:  # noqa: BLE001 - version/arch mismatches raise more than ImportError
```
drisspg: Could we narrow this exception handling? On SM100, catching every exception turns syntax errors, API breakages, and other CuTe regressions into skipped tests, so the target job can pass without collecting any of this coverage. Optional-dependency absence can be skipped, but unexpected implementation import failures should fail on supported hardware.

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    """pass if NRMSE < nrmse_tol OR max-abs-error < abs_tol.
+    ``oracle`` is the fp64 PyTorch reference; ``actual`` is the CuTe output.
+    """
+    a = actual.to(torch.float64)
+    o = oracle.to(torch.float64)
+    diff = (a - o).abs()
+    max_abs = diff.max().item()
+    rms = o.pow(2).mean().sqrt().item()
+    nrmse = diff.pow(2).mean().sqrt().item() / (rms + 1e-12)
+    assert nrmse < nrmse_tol or max_abs < abs_tol, (
```
drisspg: I think this `or` is too permissive for these large outputs: a small aggregate NRMSE can hide a severe localized error, while the absolute branch is mainly useful when the oracle is near zero. Could we always require a finite/pointwise max-error bound and use the absolute tolerance only to stabilize the relative check near zero?

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+        BT=64,
+        BC=BC,
+        BK=triton.next_power_of_2(K),
+        USE_GATHER=False,
+        CAUSAL_NORMREF=True,
+        GRID_NT=num_chunks,
+        MAX_NT=num_chunks,
+    )
+    tag = f"T={T} H={H} K={K}"
+    assert_golden(Aqk, gAqk, rAqk, dtype, f"fwd_intra Aqk {tag}")
```
drisspg: This one is all me -> we shoudl do htis ignore; in convo with the above

Could we add a prominent `# Note: [Causality Break]` around `CAUSAL_NORMREF` and reconsider its default? With `False`, the kernel uses the midpoint gate as the normalization reference, so calculations for earlier tokens read a later token. Although that reference cancels algebraically in exact arithmetic, it does not fully cancel operationally in finite precision: the later token’s magnitude can affect rounding, overflow/underflow, and exceptional-value propagation, potentially introducing future-dependent numerical bias. I think the safer default should be `True` (`normref_idx=0`), which never looks ahead. If production intentionally chooses the midpoint for numerical range, the comment should clearly document this causality tradeoff, and the test should cover both specializations.

test/test_kda_kernels.py
```diff
@@ -1026,3 +1039,1033 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    a = actual.to(torch.float64)
+    o = oracle.to(torch.float64)
+    diff = (a - o).abs()
+    max_abs = diff.max().item()
+    rms = o.pow(2).mean().sqrt().item()
+    nrmse = diff.pow(2).mean().sqrt().item() / (rms + 1e-12)
+    assert nrmse < nrmse_tol or max_abs < abs_tol, (
+        f"{name}: nrmse {nrmse:.3e} (tol {nrmse_tol:.1e}) and "
+        f"max_abs {max_abs:.3e} (tol {abs_tol:.1e}) both exceeded (ref_rms {rms:.3e})"
+    )
```
drisspg: Was this NRMSE-or-absolute-error criterion chosen deliberately? I wonder if the measuring-stick approach used by `assert_golden` above would be more informative here: compare the low-precision CuTe result and a low-precision PyTorch reference against the same fp64 reference, then bound the kernel error relative to the error introduced by low-precision inputs/computation. That seems preferable to a fixed NRMSE threshold, which can hide localized errors on large tensors. We may still want an explicit finiteness check and pointwise bound.

drisspg: Was this NRMSE-or-absolute-error criterion chosen deliberately? Would it be more aligned with the existing tests to use the measuring-stick approach from `assert_golden` above: compare both the low-precision CuTe result and a low-precision PyTorch reference against the same fp64 reference, then bound the kernel error relative to the expected low-precision error? We may still want an explicit finiteness check and pointwise bound so localized errors cannot be hidden by the aggregate metric.

test/test_kda_kernels.py
```diff
@@ -1026,3 +1041,1047 @@ def test_cumsum_scalar_vector_view_varlen():
...
+    from attn_gym.linear.kda.bwd.cute.chunk_delta_h_bwd_v1 import blackwell_delta_h_bwd_dhu_v1
+    from attn_gym.linear.kda.bwd.cute.chunk_kda_bwd_intra import (
+        chunk_kda_bwd_intra as chunk_kda_bwd_intra_cute,
+    )
+    from attn_gym.linear.kda.bwd.cute.chunk_kda_bwd_wy_dqkg_fused import (
+        chunk_kda_bwd_wy_dqkg as chunk_kda_bwd_wy_dqkg_fused_cute,
+    )
+    from attn_gym.linear.kda.fwd.cute.chunk_kda_fwd_intra import (
+        chunk_kda_fwd_intra as chunk_kda_fwd_intra_cute,
+    )
```
drisspg: also nit lets birn gimprot ot hetop  and have mabye 1 file leveo slkip if not on the right stuff

drisspg (approved):
Also give me my nice causal note ;)

PR #262, opened 2026-08-10
Add the CuTeDSL backend for selected attention

floatingtrees:
This PR adds a CuTeDSL backend for selected attention that hits 725 TFLOPS for forward and 430 TFLOPS for backward when benchmarked with:
`CUDA_VISIBLE_DEVICES=1 python benchmarks/sparse/selected_attention_benchmark.py --backend=cute --head-dim=512 --heads=128`. 

The CuTeDSL backend requires `head_dim=512` and `num_heads=128` because the flash attention kernel only supports those dimensions. It also doesn't support attention sinks at the moment. 

Testing and benchmarking was done on `a369df707e1980fb328abcc1733e3457ec10155f` for flash attention. The backwards kernels used are not yet added on the latest release (`v2.8.3`).


attn_gym/sparse/selected_attention/impl/cute.py
```diff
@@ -0,0 +1,611 @@
...
+            sliding_window_size,
+            _prebuilt_topk_idxs,
+        ) = inputs
+        O, lse, P, RowMax, topk_idxs, _unified_kv = output
+        ctx.save_for_backward(query, O, lse, P, RowMax, topk_idxs, local_kv, sparse_kv)
+        ctx.sliding_window_size = sliding_window_size
+        ctx.local_kv_len = local_kv.shape[2]
+
+    @staticmethod
+    def backward(ctx, grad_O, grad_lse, grad_P, grad_RowMax, grad_topk, grad_ukv):
```
drisspg: These are backward-only intermediates, so let's avoid materializing unused gradients and drop the redundant `unified_kv` output:

```suggestion
        return O, lse, P, RowMax, topk_idxs

    @staticmethod
    def setup_context(ctx, inputs, output):
        (
            query,
            local_kv,
            sparse_kv,
            _kv_indices,
            _attention_sink,
            sliding_window_size,
            _prebuilt_topk_idxs,
        ) = inputs
        O, lse, P, RowMax, topk_idxs = output
        ctx.set_materialize_grads(False)
        ctx.mark_non_differentiable(lse, P, RowMax, topk_idxs)
        ctx.save_for_backward(query, O, lse, P, RowMax, topk_idxs, local_kv, sparse_kv)
        ctx.sliding_window_size = sliding_window_size
        ctx.local_kv_len = local_kv.shape[2]

    @staticmethod
    def backward(ctx, grad_O, _grad_lse, _grad_P, _grad_RowMax, _grad_topk):
```

test/test_selected_attention_cute.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def test_cute_compile_fullgraph():
+    """CuTe backend works under torch.compile(fullgraph=True)."""
+    _skip_no_sm100()
+    inputs = _make_inputs(num_topk=16, seed=123)
+
+    compiled = torch.compile(selected_attention, fullgraph=True)
+
+    with torch.inference_mode():
```
drisspg: can we remove inference mode?

attn_gym/sparse/selected_attention/impl/cute.py
```diff
@@ -0,0 +1,611 @@
...
+    if query.dtype != torch.bfloat16:
+        raise TypeError("CuTe backend requires bfloat16.")
+    if not share_kv:
+        raise ValueError("CuTe backend requires share_kv=True.")
+
+    if (attention_sink > -float("inf")).any():
+        raise NotImplementedError(
+            "CuTe backend does not fuse sink correction; "
+            "attention_sink with any value > -inf is unsupported."
+        )
```
drisspg: Could we make `attention_sink` optional at the top-level API and use `None` to mean “no sink”? Then the CuTe backend could require `attention_sink is None`, which is a static check that behaves consistently under eager and `torch.compile`, rather than inspecting a CUDA tensor and synchronizing. Eager/Triton could either handle `None` directly or materialize an all-`-inf` sink. Until CuTe implements sink support, passing a tensor to this backend should raise `NotImplementedError`.

test/test_selected_attention_cute.py
```diff
@@ -0,0 +1,388 @@
...
+import math
+
+import pytest
+import torch
+
+from attn_gym.sparse.selected_attention import selected_attention
+
+ATOL_FWD = 1e-1
+RTOL_FWD = 1e-1
+ATOL_BWD = 2e-1
```
drisspg: Do we need these fixed `atol`/`rtol` values now that we have the low-precision eager comparison against an FP64 measuring stick? That check seems more principled and should catch regressions without broad tolerances that may hide them. Could we use that helper for the forward/backward correctness cases and remove these constants, assuming it covers all relevant variants?

attn_gym/sparse/selected_attention/impl/cute.py
floatingtrees: Ok, thanks for catching this.

test/test_selected_attention_cute.py
floatingtrees: I removed it.

attn_gym/sparse/selected_attention/impl/cute.py
floatingtrees: That makes sense, I updated it.

test/test_selected_attention_cute.py
floatingtrees: I moved coverage into the FP64 test and deleted these values.

floatingtrees:
Hi Driss, I've addressed all the comments. All tests pass, and so does prek.

test/test_selected_attention_cute.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def test_cute_compile_fullgraph():
+    """CuTe backend works under torch.compile(fullgraph=True)."""
+    _skip_no_sm100()
+    inputs = _make_inputs(num_topk=16, seed=123)
+
+    compiled = torch.compile(selected_attention, fullgraph=True)
+
+    with torch.inference_mode():
```
drisspg: I think this is still under `torch.inference_mode()` despite the reply saying it was removed. Could we actually drop this context so the compile test does not suppress autograd behavior?

attn_gym/sparse/selected_attention/impl/cute.py
```diff
@@ -0,0 +1,610 @@
...
+    attention_sink: torch.Tensor,
+    doc_ids: torch.Tensor | None,
+    sliding_window_size: int,
+    share_kv: bool = True,
+) -> torch.Tensor:
+    """CuTe DSL (SM100) forward for selected attention.
+
+    torch.compile(fullgraph=True) compatible. Sink correction not applied (assumes sink ≈ 0).
+    """
+    if not torch.compiler.is_compiling():
```
drisspg: I don't think the sink feedback is fully addressed yet: this guard skips `_validate_cute_constraints` under `torch.compile`, so a non-`None` sink is silently ignored instead of raising. Could we keep the static `attention_sink is not None` rejection active during compilation and add a compiled rejection test?

actuallyah why dow eneed this?

attn_gym/sparse/selected_attention/api.py
```diff
@@ -142,18 +144,18 @@ def selected_attention(
     local_kv: Tensor,
     sparse_kv: Tensor,
     kv_indices: Tensor,
-    attention_sink: Tensor,
+    attention_sink: Tensor | None,
```
drisspg: Small follow-up on making this optional: `None` is now supported, but the argument is still required at every call site. Could we give `attention_sink` a `None` default (reordering keyword-only/default arguments if needed) so it is actually optional in the public API?

test/test_selected_attention_cute.py
```diff
@@ -0,0 +1,260 @@
...
+        "sliding_window_size": sliding_window_size,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Precision vs FP64 (measuring-stick test)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("num_topk", [16, 32, 64])
```
drisspg: The measuring-stick conversion looks good, but consolidating the old tests narrowed coverage: every case now has `seq_len == sparse_seq_len == 128` and `window + topk <= 128`. That can miss local/sparse offset-splitting bugs and never exercises the padded-256 specialization. Could we retain at least one unequal-length, larger case such as `seq_len=256`, `sparse_seq_len=128`, and `window + topk > 128`?

attn_gym/sparse/selected_attention/impl/cute.py
```diff
@@ -0,0 +1,610 @@
...
+"""
+
+from __future__ import annotations
+
+import math
+from functools import lru_cache
+
+import torch
+from cutlass.cute.runtime import from_dlpack
+from flash_attn.cute.flash_bwd_mla_dq_dqv_sm100 import dQdQvGemmKernel
```
drisspg: Stepping back, I think we should simplify this backend rather than maintain a second copy of FlashAttention's integration. If we are willing to drop `torch.compile(fullgraph=True)` support for CuTe for now, could we call FA4's public `flash_attn_func(..., gather_kv_indices=..., pack_gqa=True)` directly?

The Attention Gym-specific path would then only need to:
1. validate the supported CuTe/MLA configuration;
2. build the unified local-window + offset sparse indices, including doc masking and padding;
3. concatenate local and sparse KV and convert BHSD ↔ BSHD;
4. invoke the public FA4 function and rely on its autograd, compilation cache, workspace, stream, and backward orchestration.

That should let us delete the private FA4 kernel imports, `_compile_fwd/_bwd/_dq`, the three custom ops/fake implementations, `_SelectedAttentionCuTe`, and the manual backward/workspace logic. We should also require an FA4 version containing the `-1` sentinel backward fix (beta25+), retain eager FP64 forward/backward coverage, and explicitly document CuTe as eager-only until compile support can be added through a small upstream/public wrapper rather than duplicating FA4 internals.

drisspg (changes requested):
taking a beat to look at this; what do you think about my longer integration comment

attn_gym/sparse/selected_attention/impl/cute.py
floatingtrees: I think this makes a lot of sense. I think we can directly call FA4's public function if we drop `torch.compile(fullgraph=True)` support, and the simplifications to the library seem to make the trade off worth it. I'll get started on doing that.

test/test_selected_attention_cute.py
floatingtrees: Sorry about that; I think there might have been two of those, and I only removed one. Since we're doing eager only with the rewrite, I deleted this test for now.

attn_gym/sparse/selected_attention/impl/cute.py
floatingtrees: Thanks for catching this; I removed the torch compile related code, so I don't think this is an issue anymore.

attn_gym/sparse/selected_attention/api.py
floatingtrees: That makes sense, I added it.

test/test_selected_attention_cute.py
floatingtrees: I added some more parameters to the test and it covers these cases now.

floatingtrees:
I've updated the code to only use Flash Attention's public API, removed the compile scaffolding, and fixed the other issues too. All relevant tests pass, and prek does as well. The backend is also slightly faster now, hitting 800 TFLOPs in forward and 500 TFLOPs in backward.

drisspg (approved):
Thanks!

PR #306, opened 2026-08-15
Add [linear] extra and support public CuTeDSL releases 4.5.0+

drisspg:
## Human Note

made some tweaks to support all verisons add extras

## Agent note

The KDA/short-conv CuTeDSL stack only ran against `nvidia-cutlass-dsl==4.6.0.dev0` (a non-public dev
wheel pulled in by the flash-attn cute pin) and there was no way to install its dependencies from this
package. This adds a `[linear]` extra (`nvidia-cutlass-dsl[cu13]>=4.5.0` + `apache-tvm-ffi`) so
`pip install "attn_gym[linear]"` gets everything the linear kernels need on top of a torch nightly.

CuTeDSL 4.6.2 removed the `_name_prefix=` jit-call kwarg, which broke every public release from 4.6.2
onward with `TypeError: got an unexpected keyword argument '_name_prefix'`. The replacement,
`set_name_prefix()` on the jit wrapper, has existed since at least 4.5.0, so switching `compile_tvm_ffi`
and the ten kernel launch sites to it supports 4.5.0 through 4.7.0 (current latest) with one spelling
and no version gate. The generated names are cosmetic for us — the exported TVM-FFI symbol is always
`func` — so the rename carries no cache or ABI impact. 4.4.x predates other APIs the kernels use, hence
the `>=4.5.0` floor; no upper cap since the new API is the documented one going forward.

## Test Plan

Fresh `uv venv` (py3.13) + torch nightly cu132, sweeping public CuTeDSL releases:

```bash
uv pip install -U --pre torch --index-url https://download.pytorch.org/whl/nightly/cu132
uv pip install -e ".[linear]" pytest pytest-xdist numpy typer jsonargparse docstring-parser
# for v in 4.5.0 4.6.1 4.7.0:
uv pip install "nvidia-cutlass-dsl[cu13]==$v"
pytest test -k "kda or cute or gdn" -n 4 -q   # 443 passed on each
```

Also re-ran the same 443-test selection in the primary dev env (`nvidia-cutlass-dsl==4.6.0.dev0`):
443 passed.

### DIFF
--- a/attn_gym/_backends/cute/utils.py
+++ b/attn_gym/_backends/cute/utils.py
@@ -85,8 +85,7 @@ def compile_tvm_ffi(
 
     from cutlass import cute
 
-    # CuTeDSL 4.6.2 removed the `_name_prefix=` call kwarg; `set_name_prefix()` on
-    # the jit wrapper is the compatible spelling across 4.5.0 through 4.7.0+.
+    # Requires CuTeDSL >= 4.5.0 for `set_name_prefix` (`_name_prefix=` was removed in 4.6.2).
     # Class-based entrypoints hold the jit wrapper on their `__call__`.
     jit_wrapper = entrypoint if hasattr(entrypoint, "set_name_prefix") else entrypoint.__call__
     jit_wrapper.set_name_prefix(name)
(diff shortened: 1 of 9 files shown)


drisspg:
cc @liangel-02

PR #313, opened 2026-08-15
add cu_seqlen awareness to naive KDA recurrent impl

liangel-02:
as title

### DIFF
--- a/attn_gym/linear/kda/naive.py
+++ b/attn_gym/linear/kda/naive.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from functools import reduce
 from itertools import pairwise
 
 import torch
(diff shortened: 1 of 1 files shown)


attn_gym/linear/kda/naive.py
```diff
@@ -34,9 +37,41 @@ def naive_recurrent_kda(
         beta: scalar-per-head delta step size / write gate
         scale: query scale for q k^T (optional; default 1/sqrt(K))
         initial_state: initial recurrent state (B, H, K, V) (optional)
-        output_final_state: also return the final state (optional)
+        output_final_state: also return the final state (optional; in the compute dtype)
```
drisspg, line 40: We should align the argument order here between the two - can you check that

PR #322, opened 2026-08-17
add paged awareness to decode kernel

liangel-02:


attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -66,7 +68,9 @@ def kda_recurrent_fwd_kernel(
     m_v = o_v < V
     m_kv = m_k[:, None] & m_v[None, :]
 
-    p_state = i_n * H * K * V + i_h * K * V + o_k[:, None] * V + o_v[None, :]
+    # Paged mode addresses the state pool by slot instead of by sequence position.
+    i_state = tl.load(state_indices + i_n).to(tl.int64) if USE_STATE_INDICES else i_n
```
drisspg, line 85: Can we document the routing invariant here? For every nonempty logical sequence, `state_indices[i]` must be in `[0, num_state_slots)`, and nonempty sequences must map to distinct slots. This is injective over active sequences, since unused pool rows are expected to remain untouched.

I have been using -> # Assumption for tests we dont want sync on and we can figure out where to put the right checker kernel later

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -66,7 +68,9 @@ def kda_recurrent_fwd_kernel(
...
+    p_state = i_state * H * K * V + i_h * K * V + o_k[:, None] * V + o_v[None, :]
```
drisspg, line 86: Can we return early when `bos == eos`, before loading the state? Otherwise an empty sequence sharing a slot with a nonempty sequence can load the old state and race with the active sequence by storing that stale state back. The early return makes empty/empty and empty/nonempty collisions harmless; nonempty sequences must still use distinct slots. Please add a regression test for the empty/nonempty collision.

drisspg (approved):
two comments but overall looks good :)

PR #323, opened 2026-08-17
short conv for decode

liangel-02:
to take in state for decode case, per step just shift left from the previous inputs. i think its better to create a new kernel instead of making the existing dense one paging aware because it reduces all of the generality of the dense kernel since we only need to handle the T = 1 case

perf measurements compared w/ vllm
Screenshot 2026-08-17 at 3 27 41 PM

### DIFF
--- a/attn_gym/linear/kda/short_conv/cute.py
+++ b/attn_gym/linear/kda/short_conv/cute.py
@@ -29,7 +29,16 @@ from cutlass.cute.nvgpu import cpasync
 
 from attn_gym._backends.cute import ceildiv, compile_tvm_ffi, jit_cache, tune
 from attn_gym._backends.cute.device import upper_bound
+from attn_gym.linear.kda import ops as kda_ops
 from attn_gym.linear.kda.short_conv.activations import Activation, resolve_activation
+
+_forward_op = kda_ops.short_conv_forward_op
+_backward_op = kda_ops.short_conv_backward_op
+_decode_op = kda_ops.short_conv_decode_op
+_configured_forward_op = kda_ops.short_conv_configured_forward_op
+_configured_backward_op = kda_ops.short_conv_configured_backward_op
+_configured_backward_with_state_grad_op = kda_ops.short_conv_configured_backward_with_state_grad_op
+_configured_decode_op = kda_ops.short_conv_configured_decode_op
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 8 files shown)


attn_gym/linear/kda/short_conv/cute.py
```diff
@@ -423,6 +423,106 @@ def __call__(
...
+            taps[(None, self.width - 1)].store(
+                x_groups[((0, None), (sequence, channel_group))].load()
+            )
+
+            output_groups[((0, None), (sequence, channel_group))].store(
+                self.activation(unrolled_dot(taps, weights, 0, self.width)).to(
+                    self.dtype.cute_type
+                )
+            )
+            for row in cutlass.range_constexpr(self.width - 1):
```
drisspg, line 502: This looks unsafe when `C` is not divisible by the tuned `channels_per_thread`. Please route the default through `_compatible_config(default, channels)`, validate explicit configs with `_validate_config()`, and retain a kernel-side divisibility assertion as defense in depth. An assertion alone would reject the new `C=5/6` cases rather than selecting a safe specialization.

attn_gym/linear/kda/short_conv/cute.py
```diff
@@ -3248,9 +3466,84 @@ def causal_conv1d(
     )
 
 
+def causal_conv1d_update(
```
drisspg: im debating of specifiyg hn as causal_conv1d_ (to indicate inplace)or causal_conv1d_inplace or causal_conv1d_decode 

 

up to you though your call

attn_gym/linear/kda/short_conv/cute.py
```diff
@@ -3248,9 +3466,84 @@ def causal_conv1d(
...
+            f"on x.device, got {tuple(state_indices.shape)} of {state_indices.dtype}"
+        )
+    if torch.is_grad_enabled() and any(tensor.requires_grad for tensor in (x, weight, state)):
+        raise RuntimeError(
+            "causal_conv1d_update is inference-only and has no backward; use "
+            "causal_conv1d for training or call under torch.no_grad()"
+        )
+    if forward_config is None:
+        return _decode_op(x, weight, state, state_indices, activation=activation)
+    return _launch_decode(x, weight, state, forward_config, state_indices, activation=activation)
```
drisspg: This looks like a bug: passing `forward_config` bypasses the registered opaque op and calls `_launch_decode` directly, so the tuned path performs `data_ptr()` inspection, compilation/cache work, and the TVM-FFI call while Dynamo is tracing. Please route configured decode through a dedicated scalar-config custom op (as `causal_conv1d` does) and add a `torch.compile(fullgraph=True)` test covering an explicit schedule.

test/test_kda_short_conv_cute.py
```diff
@@ -89,16 +90,67 @@ def _packed_state_reference(
...
+    x_storage = torch.randn(tokens * channels + 1, device="cuda", dtype=torch.bfloat16)
+    weight_storage = torch.randn(channels * width + 1, device="cuda", dtype=torch.bfloat16)
+    x = x_storage[1:].view(1, tokens, channels).requires_grad_()
+    weight = weight_storage[1:].view(channels, width).requires_grad_()
+    assert x.is_contiguous() and x.data_ptr() % 16 != 0
+    assert weight.is_contiguous() and weight.data_ptr() % 16 != 0
+    return x, weight
+
+
+def _paged_inputs(tokens: int = 3, channels: int = 12, width: int = 4, slots: int = 7):
```
drisspg: we shoudl probably add this paged input funcs to the testing utisl I coudl imagien this willbe reused

test/test_kda_short_conv_cute.py
```diff
@@ -89,16 +90,67 @@ def _packed_state_reference(
...
+    """One token per sequence plus a paged history pool addressed out of order."""
+    x, weight = (
+        tensor.detach()
+        for tensor in _inputs(tokens=1, channels=channels, width=width, batch=tokens)
+    )
+    state = torch.randn(slots, width - 1, channels, device="cuda", dtype=x.dtype)
+    return x.squeeze(1), weight, state, torch.tensor([5, 1, 3], device="cuda", dtype=torch.int32)
+
+
+def _dense_conv(x, weight, activation="silu"):
```
drisspg: doesnt look needed;

attn_gym/linear/kda/short_conv/cute.py
```diff
@@ -423,6 +423,106 @@ def __call__(
...
+        )
+
+    @cute.kernel
+    def kernel(
+        self,
+        x: cute.Tensor,
+        weight: cute.Tensor,
+        output: cute.Tensor,
+        state: cute.Tensor,
+        state_indices: cute.Tensor | None,
```
drisspg, line 462: will we need to add support cu_seqlens? its fine if we wnat to do that in future pr

attn_gym/linear/kda/short_conv/cute.py
```diff
@@ -3248,9 +3466,84 @@ def causal_conv1d(
...
+    sequences, width = x.shape[0], weight.shape[1]
+    slots = sequences if state_indices is None else state.shape[0]
+    if state.ndim != 3 or state.shape != (slots, width - 1, x.shape[1]):
+        raise ValueError(
+            f"state must have shape [{slots}, {width - 1}, {x.shape[1]}], got {tuple(state.shape)}"
+        )
+    if not state.is_contiguous() or state.dtype != x.dtype:
+        raise ValueError("state must be contiguous and match x dtype")
+    if weight.dtype != x.dtype or not weight.is_contiguous():
+        raise ValueError("weight must match x dtype and be contiguous")
```
drisspg: Could we add decode-specific validation equivalent to `_validate_inputs` here? We never check `weight.shape[0] == x.shape[1]`, supported CUDA dtype, positive sequence/channel dimensions, or that `x`, `weight`, and `state` share a device. Those mismatches can currently reach the raw launcher and become out-of-bounds accesses or opaque CUDA errors.

attn_gym/linear/kda/short_conv/cute.py
liangel-02: i think not needed for decode only

attn_gym/linear/kda/short_conv/cute.py
liangel-02: reanmed to causal_conv1d_decode

attn_gym/linear/kda/short_conv/cute.py
```diff
@@ -2623,6 +2842,116 @@ def _default_forward_fake(
     return torch.empty_like(x)
 
 
+torch.library.define(
```
drisspg: I made some updates all the lib defines shoudl go in ops.py

drisspg (approved):
one comment

PR #333, opened 2026-08-18
strided paged KDA state caches

liangel-02:
Screenshot 2026-08-18 at 11 04 51 AM

Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #334
* #335
* __->__ #333

PR #335, opened 2026-08-18
optimize general recurrent kernel

liangel-02:
trying different BV and stages , setting BV and num_warps to be triton autotune params

Screenshot 2026-08-18 at 11 53 06 AM

general_recurrent_launch_heatmap

Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #334
* __->__ #335
* #333

### DIFF
--- a/attn_gym/linear/kda/fwd/triton/recurrent.py
+++ b/attn_gym/linear/kda/fwd/triton/recurrent.py
@@ -15,12 +15,11 @@ inference-only; use ``chunk_kda`` for training.
 
 from __future__ import annotations
 
-import threading
-
 import torch
 import triton
 import triton.language as tl
 
+from attn_gym._backends.triton.utils import ptr_offset
 from attn_gym.linear.kda.ops import recurrent_forward as forward
 from attn_gym.linear.kda.ops import (
     recurrent_fwd_no_state_op as _recurrent_fwd_no_state_op,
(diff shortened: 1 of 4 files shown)


attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -108,7 +139,16 @@ def kda_recurrent_fwd_kernel(
...
+        if CONTIGUOUS_FINAL_STATE:
+            p_final_state = i_n * H * K * V + i_h * K * V + o_k[:, None] * V + o_v[None, :]
+        else:
+            p_final_state = p_state
+        tl.store(ht + p_final_state, b_state, mask=m_kv)
+
+
+_fixed_recurrent_fwd_kernel = kda_recurrent_fwd_kernel.fn
+_recurrent_autotune_configs: dict[tuple[object, ...], dict[str, object]] = {}
+_recurrent_autotune_lock = threading.Lock()
```
drisspg, line 151: Could we use Triton's built-in state-restoration hook here instead of adding a second winner cache and synchronization mechanism? Since paged mode aliases `h0` and `ht` to the same state pool, adding `restore_value=["h0"]` to `@triton.autotune` should restore the pool between candidate benchmarks and then let Triton perform the final winning launch exactly once. That would also benchmark the real strided pool rather than the contiguous temporary below, and should let us remove the `.fn`/`.best_config` dependency, `_recurrent_autotune_configs`, `_recurrent_autotune_lock`, `CONTIGUOUS_FINAL_STATE`, and the manual relaunch flow. We recently added centralized tuning/cache machinery to avoid each kernel growing its own version of this, so I don't think we should introduce another local cache here.

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -53,6 +83,7 @@ def kda_recurrent_fwd_kernel(
     STORE_FINAL_STATE: tl.constexpr,
     IS_VARLEN: tl.constexpr,
     USE_STATE_INDICES: tl.constexpr,
+    CONTIGUOUS_FINAL_STATE: tl.constexpr,
```
drisspg, line 86: Could we avoid adding this autotuning-only layout mode entirely? With `restore_value=["h0"]`, paged autotuning can execute the exact real paged path (`i_state -> i_state`) and Triton will restore the aliased state pool between candidate benchmarks before launching the winner once. Then this flag and `CONTIGUOUS_FINAL_STATE` branch disappear, and the common address can remain explicit without changing its layout/codegen:

```python
p_state = ptr_offset(
    (i_state, i_h, o_k[:, None], o_v[None, :]),
    (state_batch_stride, K * V, V, 1),
)
...
tl.store(ht + p_state, b_state, mask=m_kv)
```

That benchmarks the actual strided pool and avoids introducing a third read/write-layout combination solely for tuning.

drisspg (changes requested):

PR #350, opened 2026-08-19
optimize general recurrent kernel

liangel-02:
trying different BV and stages , setting BV and num_warps to be triton autotune params

Screenshot 2026-08-18 at 11 53 06 AM

general_recurrent_launch_heatmap

### DIFF
--- a/attn_gym/linear/kda/fwd/triton/recurrent.py
+++ b/attn_gym/linear/kda/fwd/triton/recurrent.py
@@ -28,10 +28,17 @@ from attn_gym.linear.kda.ops import recurrent_fwd_op as _recurrent_fwd_op
 from attn_gym.linear.kda.ops import (
     recurrent_fwd_paged_op as _recurrent_fwd_paged_op,
 )
+from attn_gym.linear.kda.utils import autotune_cache_kwargs
 
 
-@triton.jit
-def kda_recurrent_fwd_kernel(
+def _prune_recurrent_configs(configs, _named_args, V, **_):
+    """drop value tiles wider than the padded value dimension"""
+    max_block_v = max(8, triton.next_power_of_2(V))
+    return [config for config in configs if config.kwargs["BV"] <= max_block_v]
+
+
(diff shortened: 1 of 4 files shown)


attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -123,6 +131,29 @@ def kda_recurrent_fwd_kernel(
...
+        "H",
+        "K",
+        "V",
+        "USE_INITIAL_STATE",
+        "STORE_FINAL_STATE",
+        "IS_VARLEN",
+        "USE_STATE_INDICES",
+    ],
+    prune_configs_by={"early_config_prune": _prune_recurrent_configs},
+    restore_value=["h0"],
```
drisspg: oops i think i led you slightly astray - `restore_value=["h0"]` fixes the mutation correctness issue, but Triton implements it by cloning the entire tensor before each benchmark invocation and copying it back afterward. In paged mode `h0` is the full serving state pool, which may be several GiB, so this can OOM and the clone/copy can dominate the measurements. I think we should autotune only the non-paged path. For paged execution, let's use a deterministic heuristic for now, benchmark representative serving shapes offline, and bake the resulting choices into that heuristic.

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -44,6 +51,7 @@ def kda_recurrent_fwd_kernel(
     state_indices,
     scale,
     T,
+    N: tl.constexpr,
```
drisspg: I don't think we should specialize on the sequence count. `N` is only needed for the autotune key, but declaring it `tl.constexpr` forces a separate kernel compilation for every sequence count. Could we keep `N` as a regular non-specialized scalar while retaining it in `key=[...]`? Otherwise variable request batching can cause unbounded JIT/cache growth.

drisspg (approved):

PR #351, opened 2026-08-19
Fuse KDA decode preprocessing into recurrent kernel

liangel-02:
do everything after conv in one kernel for perf, matches vllm 
also this is specialized for t = 1

Screenshot 2026-08-18 at 11 37 16 AM

we should use VK layout for decode because it benefits more and this also matches vLLM layout. VK makes K the contiguous dimension and since K is the reduction dim its more efficient.  this part is most dominant for decode
image

prefill doesn't benefit has much and the difference is marginal as seqlen increases 
image

### DIFF
--- a/attn_gym/linear/kda/api.py
+++ b/attn_gym/linear/kda/api.py
@@ -12,9 +12,10 @@ inference prefill. Use ``impl`` to select fused kernels or the eager reference.
 
 from __future__ import annotations
 
-import math
+import sys
 from functools import partial
 from numbers import Real
+from typing import Literal
 
 import torch
 
@@ -28,6 +29,20 @@ from attn_gym.linear.kda.validation import SUPPORTED_INPUT_DTYPES, validate_kda_
 from attn_gym.linear.types import Impl, resolve_impl
 
 _CHUNK_SIZE = 64
+_DECODE_GATE_TRANSFORMS = {
+    "bounded": True,
+    "softplus": False,
+}
+
+
+def _resolve_decode_gate_transform(gate_transform: str) -> bool:
+    try:
+        return _DECODE_GATE_TRANSFORMS[gate_transform]
+    except KeyError:
+        supported = ", ".join(sorted(_DECODE_GATE_TRANSFORMS))
+        raise ValueError(
+            f"gate_transform must be one of {{{supported}}}, got {gate_transform!r}"
+        ) from None
 
 
 def chunk_kda(
@@ -274,29 +289,46 @@ def recurrent_kda_decode(
     state_cache: torch.Tensor,
     state_indices: torch.Tensor,
     *,
-    lower_bound: float | None = -5.0,
+    gate_transform: Literal["bounded", "softplus"] = "bounded",
+    lower_bound: float = -5.0,
     scale: float | None = None,
+    out: torch.Tensor | None = None,
 ) -> torch.Tensor:
     """Run one-token paged KDA decode with preprocessing fused into the recurrence.
 
     Args:
-        packed_qkv: Post-convolution QKV shaped ``[B, H * (2 * K + V)]``. Channels
-            are head-interleaved: each head stores its Q row, K row, then V row.
+        packed_qkv: Post-convolution QKV shaped ``[B, H * (2 * K + V)]``. Each
+            token stores ``[Q for all heads | K for all heads | V for all heads]``;
+            within each section, head rows are contiguous.
         raw_gate: Unactivated gate shaped ``[1, B, H, K]``.
         raw_beta: Unactivated write gate shaped ``[1, B, H]``.
-        A_log: Per-head log decay parameter shaped ``[H]``.
-        dt_bias: Per-head/channel gate bias shaped ``[H, K]``.
-        state_cache: FP32 paged state pool shaped ``[num_slots, H, K, V]``. Slots
-            may have padding between them but each ``[H, K, V]`` row must be dense.
-        state_indices: Contiguous int32 slot indices shaped ``[B]``. Positive slots
-            are updated in place; non-positive entries produce zero output and leave
-            the state cache untouched.
-        lower_bound: Negative bound for ``lower_bound * sigmoid(exp(A_log) * gate)``.
-            When ``None``, use ``-exp(A_log) * softplus(gate)`` instead.
+        A_log: FP32 per-head log decay parameter shaped ``[H]``.
+        dt_bias: FP32 per-head/channel gate bias shaped ``[H, K]``.
+        state_cache: FP32 paged state pool shaped ``[num_slots, H, V, K]``. Slots
+            may have padding between them but each ``[H, V, K]`` row must be dense.
+            ``K`` must be at most 256.
+            The prefill kernel uses ``[H, K, V]``, while this decode kernel uses
+            ``[H, V, K]`` because decode benefits significantly from a contiguous
+            ``K`` dimension. A prefill-to-decode transition therefore requires a
+            transpose/layout conversion.
+        state_indices: Contiguous int32 slot indices shaped ``[B]``. Non-positive
+            indices are padding/null entries: they produce zero output and leave the
+            cache untouched. Each positive index must be in ``[1, num_slots)`` and
+            unique among active rows because duplicate in-place updates race. These
+            value constraints are caller responsibilities and are not host-validated.
+        gate_transform: Pointwise gate transform. ``"bounded"`` computes
+            ``lower_bound * sigmoid(exp(A_log) * (raw_gate + dt_bias))``;
+            ``"softplus"`` computes
+            ``-exp(A_log) * softplus(raw_gate + dt_bias)``.
+        lower_bound: Finite nonpositive bound used only by the ``"bounded"`` transform.
         scale: Query scale. Defaults to ``1 / sqrt(K)``.
+        out: Optional caller-owned contiguous output buffer shaped ``[1, B, H, V]``
+            in ``packed_qkv.dtype`` on the same device. When supplied, the kernel
+            writes into and returns this exact tensor. It must not alias any input.
 
     Returns:
-        Decode output shaped ``[1, B, H, V]`` in ``packed_qkv.dtype``.
+        Decode output shaped ``[1, B, H, V]`` in ``packed_qkv.dtype``. This is
+        ``out`` itself when a buffer is supplied.
 
     Q/K L2 normalization, gate activation, beta sigmoid, the recurrent update, and
     the output projection from recurrent state are performed in one Triton kernel.
@@ -305,8 +337,8 @@ def recurrent_kda_decode(
     if packed_qkv.ndim != 2 or packed_qkv.shape[0] < 1 or packed_qkv.stride(1) != 1:
         raise ValueError("packed_qkv must have shape [B, C] and be contiguous within each token")
     if state_cache.ndim != 4:
-        raise ValueError("state_cache must have shape [num_slots, H, K, V]")
-    num_slots, heads, key_dim, value_dim = state_cache.shape
+        raise ValueError("state_cache must have shape [num_slots, H, V, K]")
+    num_slots, heads, value_dim, key_dim = state_cache.shape
     batch = packed_qkv.shape[0]
     if num_slots < 1 or heads < 1 or key_dim < 1 or value_dim < 1:
         raise ValueError(
@@ -324,15 +356,19 @@ def recurrent_kda_decode(
         )
     if raw_beta.shape != (1, batch, heads) or raw_beta.stride(2) != 1:
         raise ValueError(f"raw_beta must have shape {(1, batch, heads)} with contiguous heads")
-    if A_log.shape != (heads,) or not A_log.is_contiguous():
-        raise ValueError(f"A_log must be contiguous with shape ({heads},)")
-    if dt_bias.shape != (heads, key_dim) or not dt_bias.is_contiguous():
-        raise ValueError(f"dt_bias must be contiguous with shape ({heads}, {key_dim})")
+    if A_log.shape != (heads,) or A_log.dtype != torch.float32 or not A_log.is_contiguous():
+        raise ValueError(f"A_log must be contiguous float32 with shape ({heads},)")
+    if (
+        dt_bias.shape != (heads, key_dim)
+        or dt_bias.dtype != torch.float32
+        or not dt_bias.is_contiguous()
+    ):
+        raise ValueError(f"dt_bias must be contiguous float32 with shape ({heads}, {key_dim})")
     if state_cache.dtype != torch.float32:
         raise TypeError("state_cache must use float32")
-    expected_state_strides = (key_dim * value_dim, value_dim, 1)
+    expected_state_strides = (value_dim * key_dim, key_dim, 1)
     if state_cache.stride()[1:] != expected_state_strides:
-        raise TypeError("state_cache must be contiguous within each [H, K, V] slot")
+        raise TypeError("state_cache must be contiguous within each [H, V, K] slot")
     if state_cache.stride(0) < heads * key_dim * value_dim:
         raise ValueError("state_cache slots must not overlap")
     if (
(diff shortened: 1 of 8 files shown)


attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,118 @@ def recurrent_kda(
...
+    raw_gate: torch.Tensor,
+    raw_beta: torch.Tensor,
+    A_log: torch.Tensor,
+    dt_bias: torch.Tensor,
+    state_cache: torch.Tensor,
+    state_indices: torch.Tensor,
+    *,
+    lower_bound: float | None = -5.0,
+    scale: float | None = None,
+) -> torch.Tensor:
```
drisspg, line 296: Could we add an optional caller-owned `out` here, shaped `[1, B, H, V]` in `packed_qkv.dtype`? `None` can preserve the current allocation behavior; when supplied, the kernel writes into and returns that exact tensor. This would let serving callers reuse a stable output buffer and keep eager/CUDA-graph replay allocation-free.

attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,118 @@ def recurrent_kda(
...
+            are head-interleaved: each head stores its Q row, K row, then V row.
+        raw_gate: Unactivated gate shaped ``[1, B, H, K]``.
+        raw_beta: Unactivated write gate shaped ``[1, B, H]``.
+        A_log: Per-head log decay parameter shaped ``[H]``.
+        dt_bias: Per-head/channel gate bias shaped ``[H, K]``.
+        state_cache: FP32 paged state pool shaped ``[num_slots, H, V, K]``. Slots
+            may have padding between them but each ``[H, V, K]`` row must be dense.
+        state_indices: Contiguous int32 slot indices shaped ``[B]``. Positive slots
+            are updated in place; non-positive entries produce zero output and leave
+            the state cache untouched.
```
drisspg: Could we document the complete indexing contract here? Non-positive indices are the intentional padding/null semantics: emit zero and leave the cache untouched. Positive indices must be in `[1, num_slots)` and unique among active rows, since duplicates race on the in-place update. I don't think we need synchronizing host validation, but callers should see the requirement.

attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,118 @@ def recurrent_kda(
...
+
+    Args:
+        packed_qkv: Post-convolution QKV shaped ``[B, H * (2 * K + V)]``. Channels
+            are head-interleaved: each head stores its Q row, K row, then V row.
+        raw_gate: Unactivated gate shaped ``[1, B, H, K]``.
+        raw_beta: Unactivated write gate shaped ``[1, B, H]``.
+        A_log: Per-head log decay parameter shaped ``[H]``.
+        dt_bias: Per-head/channel gate bias shaped ``[H, K]``.
+        state_cache: FP32 paged state pool shaped ``[num_slots, H, V, K]``. Slots
+            may have padding between them but each ``[H, V, K]`` row must be dense.
```
drisspg, line 308: Could we explicitly call out that this decode cache uses `[H, V, K]`, while `recurrent_kda` and its prefill/final state use `[H, K, V]`? The VK layout makes sense for this kernel, but the public contract should explain that a prefill-to-decode transition requires a transpose/layout conversion.

attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,118 @@ def recurrent_kda(
...
+        state_indices: Contiguous int32 slot indices shaped ``[B]``. Positive slots
+            are updated in place; non-positive entries produce zero output and leave
+            the state cache untouched.
+        lower_bound: Negative bound for ``lower_bound * sigmoid(exp(A_log) * gate)``.
+            When ``None``, use ``-exp(A_log) * softplus(gate)`` instead.
```
drisspg: Could we include `dt_bias` in the documented formula so it exactly matches the kernel? Namely `lower_bound * sigmoid(exp(A_log) * (raw_gate + dt_bias))`, or `-exp(A_log) * softplus(raw_gate + dt_bias)` when the bound is `None`.

attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,118 @@ def recurrent_kda(
...
+    state_indices: torch.Tensor,
+    *,
+    lower_bound: float | None = -5.0,
+    scale: float | None = None,
+) -> torch.Tensor:
+    """Run one-token paged KDA decode with preprocessing fused into the recurrence.
+
+    Args:
+        packed_qkv: Post-convolution QKV shaped ``[B, H * (2 * K + V)]``. Channels
+            are head-interleaved: each head stores its Q row, K row, then V row.
```
drisspg: I think this input layout is wrong for the intended post-convolution boundary. `examples/kda_training.py` naturally produces each token as `[Q for all heads | K for all heads | V for all heads]` via `qkv.view(B, T, 3, H, D).unbind(2)`, and the depthwise convolution preserves that channel order; packed varlen only changes the token dimension. vLLM and SGLang use the same channel-major contract. Could we change this API/kernel to read Q at `h*K`, K at `H*K + h*K`, and V at `2*H*K + h*V`, and update the reference test to construct that layout? Otherwise an end-to-end module needs an unnecessary rearrangement, and passing the natural post-conv buffer silently reads the wrong K/V values.

drisspg:
rebase :)

attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,157 @@ def recurrent_kda(
...
+def recurrent_kda_decode(
+    packed_qkv: torch.Tensor,
+    raw_gate: torch.Tensor,
+    raw_beta: torch.Tensor,
+    A_log: torch.Tensor,
+    dt_bias: torch.Tensor,
+    state_cache: torch.Tensor,
+    state_indices: torch.Tensor,
+    *,
+    lower_bound: float | None = -5.0,
```
drisspg: With the new gate semantics on main, could we make the transform explicit here rather than encoding a different mathematical operation through `lower_bound=None`? I think the target API should be something like `gate_transform: Literal["bounded", "softplus"] = "bounded"`, with `lower_bound` used by the bounded transform. We can keep those two names closed/built-in for now, but resolve the string through a small transform table so this can grow into a `register_gate_transform(name, ...)` mechanism like short-conv activations if another pointwise gate appears. The internal op can still lower the resolved choice to a bool/string specialization; registration and compilation would happen before graph capture.

attn_gym/linear/kda/api.py
```diff
@@ -247,4 +250,157 @@ def recurrent_kda(
...
+            f"state_cache must have nonempty dimensions, got {tuple(state_cache.shape)}"
+        )
+    expected_channels = heads * (2 * key_dim + value_dim)
+    if packed_qkv.shape[1] != expected_channels:
+        raise ValueError(
+            f"packed_qkv must have shape ({batch}, {expected_channels}), "
+            f"got {tuple(packed_qkv.shape)}"
+        )
+    if packed_qkv.stride(0) < expected_channels:
+        raise ValueError("packed_qkv token rows must not overlap")
```
drisspg: Do we need to reject overlapping rows for these read-only inputs? The kernel's explicit row strides correctly support broadcast/expanded QKV, gate, or beta rows, and `B=1` cannot overlap another logical row even when `stride(0) == 0`. These checks seem to narrow the strided-input contract without a correctness need. I would remove them unless there is a downstream ABI reason to require non-overlapping rows, and add a strided/broadcast test for the intended contract.

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -123,6 +125,81 @@ def kda_recurrent_fwd_kernel(
...
+    b_gate_input = tl.load(p_gate, mask=m_k, other=0.0).to(tl.float32)
+    b_gate_input += tl.load(dt_bias + i_h * K + o_k, mask=m_k, other=0.0).to(tl.float32)
+    b_a = tl.exp(tl.load(A_log + i_h).to(tl.float32))
+    if USE_LOWER_BOUND:
+        b_gate = lower_bound * tl.sigmoid(b_a * b_gate_input)
+    else:
+        b_softplus = tl.where(
+            b_gate_input > 20.0,
+            b_gate_input,
+            tl.log(1.0 + tl.exp(b_gate_input)),
```
drisspg: If we keep `"softplus"` as an explicit gate transform, could we use a numerically stable softplus implementation and add a negative-tail regression? This branch protects the positive tail, but `log(1 + exp(x))` cancels to exactly zero for sufficiently negative FP32 `x`, unlike the PyTorch/reference softplus. The default bounded path is unaffected, but a named softplus mode should match its documented transform across the representable range.

test/test_kda_recurrent_decode.py
```diff
@@ -0,0 +1,414 @@
...
+            state_indices,
+            out=torch.empty(
+                1,
+                2,
+                state_cache.shape[1],
+                state_cache.shape[2],
+                device="cuda",
+                dtype=torch.float32,
+            ),
+        )
```
drisspg, line 350: Since the public wrapper now adds an explicit `K <= 256` contract, could we cover the boundary here with an accepted `K=256` case and a `K=257` rejection? That keeps the launcher limitation from drifting independently of validation.

drisspg (approved):
approving for now I thinkt he main thing we need is to do the udpates since I had a bigish change on the gate semantics; since that its this it shoudl be fine but see my co-aied comments

attn_gym/linear/kda/api.py
```diff
@@ -12,19 +12,37 @@
...
 from attn_gym.linear.types import Impl, resolve_impl
 
 _CHUNK_SIZE = 64
+_DECODE_GATE_TRANSFORMS = {
+    "bounded": True,
+    "softplus": False,
+}
+
+
+def _resolve_decode_gate_transform(gate_transform: str) -> bool:
```
drisspg, line 38: I think maybe put this somewehre else?  probs in validation

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -152,6 +154,85 @@ def _kda_recurrent_fwd_kernel(
...
+    m_v = o_v < V
+    m_vk = m_v[:, None] & m_k[None, :]
+
+    i_state = tl.load(state_indices + i_n).to(tl.int64)
+    row = i_n * H + i_h
+    if i_state <= 0:
+        tl.store(output + row * V + o_v, 0.0, mask=m_v)
+        return
+
+    p_state = i_state * state_batch_stride + i_h * V * K + o_v[:, None] * K + o_k[None, :]
```
drisspg, line 201: my usual use `ptr_offset` just cause it think it keeps indexing tidyer

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -152,6 +154,85 @@ def _kda_recurrent_fwd_kernel(
 )(_kda_recurrent_fwd_kernel)
 
 
+@triton.jit
+def kda_recurrent_decode_kernel(
```
drisspg, line 161: nit; maybe a quick note on thestrategy byehind this kerenel; how its parallelized

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -152,6 +154,85 @@ def _kda_recurrent_fwd_kernel(
...
+    state_batch_stride: tl.constexpr,
+    qkv_token_stride: tl.constexpr,
+    gate_token_stride: tl.constexpr,
+    beta_token_stride: tl.constexpr,
+    H: tl.constexpr,
+    K: tl.constexpr,
+    V: tl.constexpr,
+    BK: tl.constexpr,
+    BV: tl.constexpr,
+    USE_LOWER_BOUND: tl.constexpr,
```
drisspg, line 181: you should pass in stride tuples and make as minimum assumptions as possible; e.g. see the `last_dim_contiguous` stuff I added recetnlyish

test/test_kda_recurrent_decode.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+def test_recurrent_decode_cuda_graph_replay():
+    inputs = _decode_inputs(batch=3, seed=4)
+    packed_qkv, raw_gate, raw_beta, A_log, dt_bias, storage, state_cache, state_indices = inputs
+    out = packed_qkv.new_empty(1, 3, state_cache.shape[1], state_cache.shape[2])
+    recurrent_kda_decode(
+        packed_qkv, raw_gate, raw_beta, A_log, dt_bias, state_cache, state_indices, out=out
+    )
+    torch.cuda.synchronize()
```
drisspg, line 524: general note; Do a once over on all these tets make sure hey are minimum and have claude to `simplify make faster and maintian exact coverage` just to keepp ci fast

attn_gym/linear/kda/fwd/triton/recurrent.py
```diff
@@ -152,6 +154,88 @@ def _kda_recurrent_fwd_kernel(
...
+
+    p_gate = raw_gate + i_n * gate_token_stride + i_h * K + o_k
+    b_gate_input = tl.load(p_gate, mask=m_k, other=0.0).to(tl.float32)
+    b_gate_input += tl.load(dt_bias + i_h * K + o_k, mask=m_k, other=0.0).to(tl.float32)
+    b_a = tl.exp(tl.load(A_log + i_h).to(tl.float32))
+    if USE_LOWER_BOUND:
+        b_gate = lower_bound * tl.sigmoid(b_a * b_gate_input)
+    else:
+        b_softplus = tl.where(
+            b_gate_input > 20.0,
```
drisspg, line 219: I think the negative-tail fix needs a higher crossover or a fully stable formulation. FP32 cancellation in `log(1 + exp(x))` starts around `x ~= -16.6`, so values such as `x=-18` still take this middle branch and produce exactly zero. With a large positive `A_log`, that can turn a real decay into no decay. Could we use the `exp(x)` approximation from a safer threshold (for example `-10`) and add a focused reference test around `raw_gate + dt_bias = -18`?

attn_gym/linear/kda/api.py
```diff
@@ -262,4 +280,153 @@ def recurrent_kda(
...
+    ):
+        raise ValueError(f"state_indices must be contiguous int32 with shape ({batch},)")
+    if key_dim > 256:
+        raise ValueError(f"recurrent_kda_decode requires K in [1, 256], got {key_dim}")
+
+    device = packed_qkv.device
+    data_tensors = (packed_qkv, raw_gate, raw_beta, A_log, dt_bias)
+    if any(tensor.device != device for tensor in (*data_tensors, state_cache, state_indices)):
+        raise ValueError("all recurrent_kda_decode inputs must be on the same device")
+    if any(tensor.dtype not in SUPPORTED_INPUT_DTYPES for tensor in data_tensors):
```
drisspg: Could we align these parameter dtypes with the new public `bound_gate` contract? `bound_gate` requires FP32 `A_log` and `dt_bias`, while this currently accepts FP16/BF16 for both and only casts after loading, so the supposedly equivalent bounded transform can use rounded parameters in decode. I think QKV/raw gate/raw beta can keep the supported activation dtypes, but `A_log` and `dt_bias` should be validated as FP32 with rejection coverage.

attn_gym/linear/kda/api.py
```diff
@@ -262,4 +280,153 @@ def recurrent_kda(
...
+        supported = ", ".join(str(dtype) for dtype in SUPPORTED_INPUT_DTYPES)
+        raise TypeError(f"decode data inputs must use one of {supported}")
+
+    use_lower_bound = _resolve_decode_gate_transform(gate_transform)
+    if not isinstance(lower_bound, Real) or isinstance(lower_bound, bool):
+        raise TypeError("lower_bound must be a real scalar")
+    lower_bound = float(lower_bound)
+    if not math.isfinite(lower_bound) or lower_bound >= 0:
```
drisspg: Can this validation depend on `gate_transform` and match `bound_gate`? For `"bounded"`, main now accepts any finite nonpositive bound, including `0`; for `"softplus"`, `lower_bound` is unused and should not determine whether the call is valid. As written, decode rejects a valid bounded zero and also rejects NaN/positive `lower_bound` in softplus mode even though the kernel ignores it.

drisspg (commented):
soem more ai comments that look reasonable

PR #354, opened 2026-08-20
paging awareness + state_indices for chunked kernel

liangel-02:
as title, we need paging awarness for the chunked kernels also need to pass in state_indices so we can write direclty to the correct place 

this is using KV layout inside the kernel but when we write we need to be aware that decode is VK to match vllm and for performance reasons

### DIFF
--- a/attn_gym/linear/kda/api.py
+++ b/attn_gym/linear/kda/api.py
@@ -6,8 +6,9 @@
 
 """Public KDA operations.
 
-``chunk_kda`` supports training and prefill; ``recurrent_kda`` supports decode and
-inference prefill. Use ``impl`` to select fused kernels or the eager reference.
+``chunk_kda`` supports training and prefill; ``paged_chunk_kda`` advances a mutable
+state cache during inference; ``recurrent_kda`` supports decode and inference prefill.
+Use ``impl`` to select fused kernels or the eager reference where available.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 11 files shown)


attn_gym/linear/kda/api.py
```diff
@@ -36,6 +36,8 @@ def chunk_kda(
     *,
     cu_seqlens: torch.Tensor | None = None,
     output_final_state: bool = False,
+    state_indices: torch.Tensor | None = None,
+    has_initial_state: torch.Tensor | None = None,
```
drisspg: hmmmmm

attn_gym/linear/kda/api.py
```diff
@@ -57,6 +59,15 @@ def chunk_kda(
...
             and may end before ``T``.
         output_final_state: Return the final recurrent state with the output.
+            Rejected together with ``state_indices``, which advances a paged
+            state pool in place.
+        state_indices: Contiguous ``int32`` slot indices, one per logical sequence,
+            selecting rows of a paged ``initial_state`` pool shaped
+            ``[num_slots, H, V, K]``. Positive slots are advanced in place;
+            non-positive slots produce zero output and leave the pool untouched.
+            ``"fused"`` inference only.
+        has_initial_state: Optional contiguous boolean mask, one per logical sequence.
```
drisspg: we can disucss this; but I think jamming this paged form in feels bad; if it was just this state_indices I thikn its cohereint bu this has_initial-state is weird a shit to me; maybe we can brianstorm why thisis neede

Could we expose the paged form as a separate public API (for example, `paged_chunk_kda`)? Paged state has substantially different semantics: `initial_state` becomes a mutable pool, `state_indices` is required, `has_initial_state` only applies there, and the pool is advanced in place. The regular `chunk_kda(initial_state=..., output_final_state=True)` path would remain useful for cases such as context parallelism, where state is passed explicitly between context partitions.

drisspg (approved):

liangel-02:
as a note: we are keeping has_initial_state bc of the perf hit of zeroing out the cache 
image

PR #358, opened 2026-08-22
Select persistent KDA inter-solve after three waves

drisspg:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #359
* __->__ #358

## Agent note

The shared four-wave AUTO threshold keeps ragged K3/K4 on their capacity-sized static grids at the
`T=2048`, `N=512` CUDA Graph shape even though their existing persistent implementations are much
faster for sparse replays. Give the composed inter-solve a local three-wave crossover while leaving
the scheduler default and every other KDA stage unchanged. K3 and K4 continue to share one resolved
schedule.

## Performance

B200, BF16 K3+K4 CUDA Graph replay, `T=2048`, `N=512`, `M=32`, `H=16`:

| schedule | median |
|---|---:|
| static | 223.84 us |
| persistent | 40.88 us |
| AUTO | 40.91 us |

Across `M=1,8,32,64,128,256,384,512`, persistent ranged from 5.44x faster at `M=32` to 1.9%
slower at the fully active `M=512` endpoint.

## Test Plan

```bash
gpu-run auto -- env PYTHONPATH=$PWD .venv/bin/pytest test/test_kda_ragged_k3.py test/test_kda_ragged_k4.py -n 6 -q
gpu-run auto -- env PYTHONPATH=$PWD .venv/bin/python /home/dev/meta/attention-gym/agent_space/cuda_graph_overcapture/benchmark_k3_k4_schedule.py --output agent_space/k3_k4_selector_after.json --samples 30
```

PR #359, opened 2026-08-22
Bound ragged KDA dHU work by active sequences

drisspg:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* __->__ #359
* #358

## Agent note

The ragged dHU kernel already launches one fixed SM-sized grid, but every worker strides through the
captured `N × heads × value_tiles` sequence space. Derive the trailing active sequence extent from
device `cu_seqlens` and use it as the persistent loop bound so CUDA Graph replay skips repeated-end
sequence slots without host synchronization.

Compile this path only in the measured graph-overcapture region and only when no initial-state
gradient is requested. Calls with `h0` retain the original `N`-slot specialization so empty-sequence
`dh0 = dht` semantics and the existing code generation remain unchanged. A warp-local lane-zero
lookup and shuffle give every pipeline role the same trip count without adding a CTA barrier.

## Performance

B200, BF16 dHU CUDA Graph replay, `T=2048`, `N=512`, `H=16`, no initial state:

| M | before | after | result |
|---:|---:|---:|---:|
| 1 | 111.94 us | 50.05 us | 2.24x |
| 32 | 92.13 us | 35.23 us | 2.62x |
| 512 | 1,486.21 us | 1,577.38 us | 0.94x |

On top of the K3/K4 selector layer, full public `chunk_kda` forward/backward replay moves from
495.26 to 437.63 us at `M=32` (1.13x) and from 5,205.79 to 5,284.45 us at `M=512` (1.5% slower).

## Test Plan

```bash
gpu-run auto -- env CUTE_DSL_NO_CACHE=1 PYTHONPATH=$PWD .venv/bin/pytest test/test_kda_ragged_delta_h_bwd.py -q
gpu-run auto -- env PYTHONPATH=$PWD .venv/bin/pytest -n 6 -q test/test_kda_ragged_k3.py test/test_kda_ragged_k4.py test/test_kda_ragged_delta_h_bwd.py test/test_kda_ragged_public_backward.py test/test_kda_cute_forward.py test/test_kda_int64_offsets.py test/test_kda_compile_dynamic_shapes.py test/test_kda_chunk_scheduler.py
gpu-run auto -- env CUTE_DSL_NO_CACHE=1 PYTHONPATH=$PWD .venv/bin/python agent_space/dhu_m_sweep.py --label final --output agent_space/dhu_final.json --samples 15
```

### DIFF
--- a/attn_gym/linear/kda/bwd/cute/chunk_delta_h_bwd_v1.py
+++ b/attn_gym/linear/kda/bwd/cute/chunk_delta_h_bwd_v1.py
@@ -39,6 +39,7 @@ from attn_gym._backends.cute.cache import jit_cache
 from attn_gym._backends.cute.target import get_compile_target
 from attn_gym._backends.cute.utils import compile_tvm_ffi, requires_int64_abi
 from attn_gym.linear.kda.chunk_scheduler import RaggedChunkMetadata
+from attn_gym.linear.kda.fwd.cute.chunk_scheduler_cute import load_ragged_sequence_extent
 
 # ============================================================================
 # BlackwellDeltaHBwdV1 — warp-specialized backward inter-chunk recurrence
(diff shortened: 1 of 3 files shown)


drisspg:
Superseded by the GitHub-native stacked PRs requested for this repository.

PR #375, opened 2026-08-26
Add native FP16 KDA training

drisspg:
## Agent note

Keep homogeneous FP16 Q/K/V in FP16 across the composed KDA forward and backward instead of
normalizing the complete private ABI to BF16. The FP16 forward uses the FP32/TF32 diagonal factor
path where strip-rebased FP16 operands can overflow; FP32 and mixed-dtype callers retain the
existing BF16 normalization.

The backward kernels select FP16 tcgen05 descriptors and use native CuTeDSL/TensorSSA for ordinary
conversion and synchronization operations. The WY kernel also drops roughly 1,500 lines of unused
masked/TF32/TS helper-library surface while preserving the source attribution.

FP16 has less exponent range than BF16. The documentation recommends L2-normalized or otherwise
bounded Q/K, as used by the training example: intermediate GEMMs accumulate in FP32, but their
results are converted back to FP16 when consumed by the next GEMM.

## Performance

B200 public `chunk_kda` forward plus autograd, fixed warm tensors, CUDA Graph replay,
`autotune=False`:

| Workload | BF16-bridge baseline | Native FP16 | Change |
| --- | ---: | ---: | ---: |
| T1024/H16 | 248.039 us | 194.274 us | -21.7% |
| packed `[65,1024,4096,0,127]`/H16 | 1128.815 us | 912.489 us | -19.2% |
| T8192/H64 | 4969.449 us | 3729.254 us | -25.0% |

BF16 T8192/H64 remained stable: 3572.415 us before and 3570.097 us after (-0.065%).

## Test Plan

```bash
pytest -n 6 -q  # 974 passed, 1 skipped
prek --all-files
mkdocs build --strict

gpu-run auto -- cutracer trace \
  --cutracer-so ~/.cache/cutracer/current/lib/cutracer.so \
  -a random_delay --delay-ns 20000 --delay-enable-prob 1.0 \
  --delay-dump-path agent_space/cutracer_fp16_wy_stress/delay.json \
  -k ChunkKdaBwdWyDqkgFused -- \
  pytest -q 'test/test_kda_ragged_bwd_wy.py::test_ragged_wy_matches_independent_numerical_reference[dtype1]'
```

### DIFF
--- a/attn_gym/linear/kda/bwd/cute/chunk_delta_h_bwd.py
+++ b/attn_gym/linear/kda/bwd/cute/chunk_delta_h_bwd.py
@@ -53,6 +53,14 @@ from attn_gym.utils import ceildiv
 
 _MIN_SEQUENCE_EXTENT_SEQUENCES = 32
 _MIN_SEQUENCE_EXTENT_HEADS = 8
+_IO_TYPES = {
+    torch.float16: cutlass.Float16,
+    torch.bfloat16: cutlass.BFloat16,
+}
+_IO_TYPE_NAMES = {
+    cutlass.Float16: "fp16",
+    cutlass.BFloat16: "bf16",
+}
 
 
 def select_delta_h_bv(
@@ -185,6 +193,7 @@ class BlackwellDeltaHBwd:
     def __init__(
         self,
         num_heads: int,
+        io_type: type[cutlass.Numeric] = cutlass.BFloat16,
         head_bv: int = 16,
         varlen: bool = False,
         use_int64_offsets: bool = False,
@@ -194,7 +203,7 @@ class BlackwellDeltaHBwd:
         self.head_k = 128
         self.head_v = 128
         self.acc_type = cutlass.Float32
-        self.io_type = cutlass.BFloat16
+        self.io_type = io_type
         self.num_heads = num_heads
         self.varlen = varlen
         self.use_int64_offsets = use_int64_offsets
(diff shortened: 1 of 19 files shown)


attn_gym/linear/kda/bwd/cute/chunk_delta_h_bwd.py
```diff
@@ -194,7 +203,7 @@ def __init__(
         self.head_k = 128
         self.head_v = 128
```
drisspg, line 204: just kep thename look internal?

attn_gym/linear/kda/bwd/cute/chunk_delta_h_bwd.py
```diff
@@ -2216,6 +2226,7 @@ def _compile_delta_h_bwd(H, bv, use_int64_offsets):
 def _compile_delta_h_bwd_packed(
     heads: int,
     bv: int,
```
drisspg, line 2228: dont ahve duplciate name and type just map from type toname

attn_gym/linear/kda/bwd/cute/chunk_kda_bwd_intra.py
```diff
@@ -2025,6 +1936,7 @@ def _compile_chunk_kda_bwd_intra(
     capacity: int,
```
drisspg, line 1936: same herre

PR #378, opened 2026-08-26
Get Selected Attention to return LSE for auxilary loss computation 

floatingtrees:
Updated Selected Attention to return LSE computed over all branches + sink. The CSA example now has a function that computes auxiliary indexer loss with the Deepseek formulation. All relevant tests pass.


attn_gym/sparse/selected_attention/impl/triton/__init__.py
```diff
@@ -50,10 +50,11 @@ def forward(
         ctx.doc_ids = doc_ids
         ctx.sliding_window_size = sliding_window_size
         ctx.share_kv = share_kv
-        return output
+        ctx.mark_non_differentiable(lse)
```
drisspg, line 53: we could makes this differentiable thouggh right?

attn_gym/sparse/selected_attention/impl/triton/__init__.py
```diff
@@ -20,7 +20,7 @@ def forward(
         doc_ids: torch.Tensor | None,
         sliding_window_size: int,
         share_kv: bool,
-    ) -> torch.Tensor:
+    ) -> tuple[torch.Tensor, torch.Tensor]:
```
drisspg, line 23: also in pytorch at leat for the main interface we have this return_aux that users can say if htey wnt lse out of the top levl api

attn_gym/sparse/selected_attention/impl/triton/__init__.py
floatingtrees: It's possible, but flash MLA doesn't return dLSE. The current auxiliary loss also doesn't use it, so I'm not sure if it's worth the effort.

attn_gym/sparse/selected_attention/impl/triton/__init__.py
floatingtrees: Ok, I'll add that

floatingtrees:
I fixed the `return_aux` issue, and all relevant tests pass.

attn_gym/sparse/selected_attention/impl/triton/__init__.py
floatingtrees: What do you think?

examples/compressed_sparse_attention.py
```diff
@@ -367,6 +370,64 @@ def CSA(
...
+
+    indexer_probs = F.softmax(selected_indexer_logits.float(), dim=-1)
+    # Return KL
+    return (
+        (
+            compressed_teacher_probs
+            * (compressed_teacher_probs.clamp_min(eps).log() - indexer_probs.clamp_min(eps).log())
+        )
+        .sum(dim=-1)
+        .mean()
```
drisspg: A safe softmax would avoid the NaN for an all-masked row, but I think we still need the validity mask here: those prefix rows have no valid sparse target and should be excluded from the loss, rather than treated as a probability distribution. Could we carry `selected_is_valid` into `indexer_loss`, mask invalid selections, exclude all-invalid rows from the reduction, and add a test covering the first `compression_rate - 1` positions?

examples/compressed_sparse_attention.py
```diff
@@ -367,6 +370,64 @@ def CSA(
...
+    softmax_scale = head_dim**-0.5
+
+    # Recompute main-attention logits for selected compressed keys.
+    # Selected attention doesn't store an attention matrix, so we have to do this
+    compressed_attention_logits = (
+        torch.einsum(
+            "bhsd,bskd->bhsk",
+            main_query.detach(),
+            selected_compressed_kv.detach(),
+        )
```
drisspg, line 408: Could we make `indexer_loss` handle CSA’s actual multi-head compressed keys and add an end-to-end multi-head test? `compress` returns `[B, H, blocks, D]` here (including when shared inputs are expanded before applying per-head biases), so gathering the selected keys produces `[B, H, S, K, D]`. The current `bskd` einsum expects `[B, S, K, D]` and raises a rank error; this should be `bhsd,bhskd->bhsk`.

examples/compressed_sparse_attention.py
```diff
@@ -367,6 +374,78 @@ def CSA(
...
+
+    masked_logits = selected_indexer_logits.float().masked_fill(~selected_is_valid, float("-inf"))
+    row_has_valid = selected_is_valid.any(dim=-1)
+    masked_logits = torch.where(
+        row_has_valid[..., None],
+        masked_logits,
+        torch.zeros_like(masked_logits),
+    )
+    indexer_probs = F.softmax(masked_logits, dim=-1)
+
```
drisspg, line 435: Could we make this reduction branchless? `row_has_valid.any()` is data-dependent Python control flow, so this fails under `torch.compile(..., fullgraph=True)` and requires a device-to-host scalar read on CUDA. The all-invalid return is also detached from `selected_indexer_logits`, so a standalone `loss.backward()` fails instead of producing zero gradients.

Please first make all-invalid rows safe before `softmax` (the current all-`-inf` logits produce NaNs), e.g. use `torch.where(row_has_valid[..., None], masked_logits, 0)`, then reduce with something like `valid_kl.sum() / row_has_valid.sum().clamp_min(1)`, masking invalid `kl` rows with `torch.where`. Could we also cover mixed-validity and all-invalid eager backward plus `torch.compile(fullgraph=True)` forward/backward?

examples/compressed_sparse_attention.py
```diff
@@ -367,6 +374,78 @@ def CSA(
...
+    softmax_scale = head_dim**-0.5
+
+    # Recompute main-attention logits for selected compressed keys.
+    # Selected attention doesn't store an attention matrix, so we have to do this
+    compressed_attention_logits = (
+        torch.einsum(
+            "bhsd,bhskd->bhsk",
+            main_query.detach(),
+            selected_compressed_kv.detach(),
+        )
```
drisspg, line 408: Would be interesting to verify this for BF16/FP16. Selected attention accumulates the QK dot in FP32, while this `einsum` appears to produce low-precision logits before combining them with the FP32 LSE. Should we cast the detached query and KV to FP32 here so the teacher probabilities match the attention computation?

examples/compressed_sparse_attention.py
```diff
@@ -367,6 +374,78 @@ def CSA(
     )
 
 
+def indexer_loss(
```
drisspg, line 377: Could the end-to-end test use `num_topk_blocks >= 2` and check the loss/gradient against an oracle? With one selected block both distributions are always `[1]`, so the KL and its gradient are always zero and the test can pass without validating the indexer-loss behavior.

examples/compressed_sparse_attention.py
floatingtrees: Thanks for catching this, I fixed it.

examples/compressed_sparse_attention.py
floatingtrees: Ok, I updated the code for this too.

drisspg (approved):
LGTM thanks

drisspg:
Just fix lint and shoudl be good to land

floatingtrees:
Linting and prek pass now.

PR #389, opened 2026-08-27
Add recurrent_gdn_decode with fused QK L2 normalization

drisspg:
Stacked PRs:
 * #390
 * __->__#389
 * #388

--- --- ---

Add recurrent_gdn_decode with fused QK L2 normalization

## Agent note

Add the serving-specific one-token GDN entrypoint mirroring recurrent_kda_decode: 3-D per-token
inputs, the paged FP32 state pool advanced in place through state_indices with has_initial_state
fresh-slot semantics, grouped q/k heads, and the per-token query/key L2 normalization fused into
the scan registers (x / sqrt(sum(x^2) + 1e-6), bit-matching FLA use_qk_l2norm_in_kernel) so decode
callers launch no separate elementwise kernels. The shared scan gains a QK_L2NORM constexpr; the
op registers as attn_gym::gdn_recurrent_decode with a mutating schema, fake registration, opcheck,
fullgraph, and CUDA-graph coverage. Exact-equality tests build identically strided pools because
state_batch_stride is a kernel constexpr and different pool layouts compile different FP schedules.

## Test Plan

```bash
.venv/bin/prek run --files attn_gym/linear/_delta_rule/recurrent.py attn_gym/linear/gdn/ops.py attn_gym/linear/gdn/api.py attn_gym/linear/gdn/impl/fused.py test/linear/test_gdn_decode.py docs/linear.md
gpu-run auto -- .venv/bin/pytest -q -n 6 test/linear/ test/test_kda_recurrent.py test/test_kda_recurrent_decode.py test/test_namespaces.py
```

### DIFF
--- a/attn_gym/linear/__init__.py
+++ b/attn_gym/linear/__init__.py
@@ -9,7 +9,7 @@
 import importlib
 from typing import TYPE_CHECKING
 
-from attn_gym.linear.gdn import chunk_gdn, recurrent_gdn
+from attn_gym.linear.gdn import chunk_gdn, recurrent_gdn, recurrent_gdn_decode
 from attn_gym.linear.kda import (
     active_token_mask,
     chunk_kda,
@@ -42,6 +42,7 @@ KDA_OPS = [
 GDN_OPS = [
     "chunk_gdn",
     "recurrent_gdn",
+    "recurrent_gdn_decode",
 ]
 
 # Model-agnostic building blocks; they currently ship from the KDA module.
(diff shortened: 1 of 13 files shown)


drisspg:
Pulling this out of the stack: the decode entrypoint is being rebuilt as a generalization of the dedicated KDA decode kernel (packed-QKV ABI, in-kernel gate transform, GQA) rather than a flag on the shared scan. #388 and #390 land first; the replacement will supersede this PR.

PR #394, opened 2026-08-27
Add fused chunk_gdn over the KDA chunk pipeline

drisspg:
Stacked PRs:
 * #395
 * __->__#394

--- --- ---

Add fused chunk_gdn over the KDA chunk pipeline

## Agent note

Implement chunk_gdn(impl="fused") as an adapter over the fused KDA chunk pipeline: GDN's
recurrence is KDA's with one decay shared by every key channel, so expanding the scalar gate per
channel (and repeat_interleave-ing grouped q/k heads) computes it exactly on the existing CuTe
kernels. The expansions stay inside autograd, whose backward reduces the gate gradient over K and
the q/k gradients over each value-head group, so training works end to end with no new kernels.
Scale overrides fold into q since the core fixes 1/sqrt(K). Requires CUDA capability 10.0 and
K = V = 128, enforced by the shared core validation; CPU calls raise before backend import.

Tests pin forward (dense, grouped, initial-state, packed, scale override) and full backward
gradients against the eager reference with FP64 measuring-stick budgets. GB200, B4/Hk4/Hv12/128:
forward 181.7 us at T1024 and 676.2 us at T4096 vs 142.8/536.2 us for the raw equal-heads
chunk_kda core -- the ~26% adapter overhead (gate expansion + K-redundant cumsum) is the headroom
a scalar-gate specialization of the intra stage can later reclaim.

## Test Plan

```bash
.venv/bin/prek run --files attn_gym/linear/gdn/impl/fused.py attn_gym/linear/gdn/ops.py attn_gym/linear/gdn/api.py attn_gym/testing/__init__.py test/linear/test_gdn.py test/linear/test_gdn_chunk_fused.py
gpu-run auto -- .venv/bin/pytest -q test/linear/test_gdn_chunk_fused.py
gpu-run auto -- .venv/bin/pytest -q -n 6 test/linear/ test/test_kda_recurrent.py test/test_kda_cute_forward.py
gpu-run auto -- .venv/bin/python agent_space/bench_chunk_gdn.py adapter
```

### DIFF
--- a/attn_gym/linear/_delta_rule/validation.py
+++ b/attn_gym/linear/_delta_rule/validation.py
@@ -2,9 +2,20 @@
 
 from __future__ import annotations
 
+from numbers import Real
+
 import torch
 
 SUPPORTED_ACTIVATION_DTYPES = (torch.float16, torch.bfloat16, torch.float32)
+
+
+def resolve_scale(scale: float | None, key_dim: int) -> float:
+    """Resolve a query-scale override to a concrete float, defaulting to ``1/sqrt(K)``."""
+    if scale is None:
+        return key_dim**-0.5
+    if not isinstance(scale, Real) or isinstance(scale, bool):
+        raise TypeError("scale must be a real scalar or None")
+    return float(scale)
 
 
 def validate_has_initial_state(
@@ -148,6 +159,7 @@ def validate_paged_state(
 __all__ = [
     "SUPPORTED_ACTIVATION_DTYPES",
     "resolve_decode_out",
+    "resolve_scale",
     "validate_decode_inputs",
     "validate_has_initial_state",
     "validate_paged_state",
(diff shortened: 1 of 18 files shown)


drisspg:
Dropping this in favor of waiting for the Mega chunk backend: the fused chunk pipeline's intra-chunk factorization overflows FP32 once any per-token gate drops below ~-5.914 (128 / (15 * log2(e)), see kda/constants.py), and GDN's -exp(A_log)*softplus(...) gate is unbounded below — a Qwen3-Next-style init reaches ~-20 on its largest heads, so realistic GDN training/prefill would produce non-finite outputs. KDA is only safe because its bounded gate transform enforces the limit contractually. The Mega intra stage will use an underflow-only rebase and the adapter will return on top of it (work preserved on backup/chunk-gdn-adapter). The host-driven chunk_kda scale generalization from this PR moved to #398.

PR #421, opened 2026-08-30
Add fla backend for KDA and KDA Context Parallelism (KCP)

QIU023:
### Summary

Adds `Impl.FLA`: `chunk_kda` over flash-linear-attention's triton kernels, sitting between the SM100-only fused path and the pure-PyTorch reference -- arch-agnostic like reference, kernel-grade like fused. This picks up the "dispatch to FLA on unsupported archs" direction from the kda CP example discussion; `fla` is an optional dependency (importing the backend without it raises with an install pointer), the same standing it has in torchtitan's qwen3_5 today. A second commit exposes fla's own context-parallel machinery behind this package's conventions, so KDA runs context-parallel on any CUDA architecture.

### Design

- `Impl` gains `FLA`; the `chunk_kda` dispatch gains one branch; the adapter lives in `kda/fla_backend.py`.
  - The adapter feeds fla the exact contract this package exposes -- q/k pre-normalized, `gate` the bound natural-log decay, `beta` already sigmoided -- so every `use_*_in_kernel` switch stays off. No convention translation is needed except one: the two libraries transpose the recurrent state's last two axes relative to each other, and the adapter owns that flip in both directions (`initial_state` in, `final_state` out).
- `kda/fla_cp.py` wraps fla's CP mechanism: `build_fla_cp_context` (the delta recurrence becomes a prefix scan over rank-local fragments; `cu_seqlens` carries global packed-document boundaries, defaulting to one document), `causal_conv1d_cp` (the short-convolution halo exchange), and `chunk_kda_cp` (this rank's fragment through the scan).
  - `initial_state`/`output_final_state` are unsupported under CP: the final state only matters for decoding, which does not run context-parallel.
- `bound_gate` is unchanged: the fla backend consumes its reference-impl output.

### Results

FLA against REFERENCE on the same bf16 inputs (B=2, T=128, H=4, K=V=128, RTX 5060 Ti / SM120 -- no Blackwell-DC hardware involved):

| check | max diff |
|---|---|
| forward output | 2.4e-4 |
| final state (after the axis flip; 0.61 before it) | 2.9e-3 |
| FLA half-run resumed from its own final state vs one full run | 0.0 (bitwise) |
| REFERENCE final state resumed by FLA (cross-impl handoff) | 1.2e-4 |
| gradients dq/dk/dv/dgate/dbeta | <= 7.5e-9 |
| varlen, two documents via `cu_seqlens` | 2.4e-4 |

Context parallel, two ranks over contiguous shards of T=256 against the single-rank run:

| check | max diff |
|---|---|
| forward, each rank's fragment vs the full run's slice | 6.1e-5 |
| backward through the cross-rank scan | gradients finite, norms match |

The cross-impl state handoff means CP orchestration built on state composition can run its per-rank compute on FLA without the composition layer noticing.

### Changed files

    attn_gym/linear/
      types.py           +1     Impl.FLA
      kda/api.py         +14    the dispatch branch
      kda/fla_backend.py +81    the adapter and the state-axis flip (new)
      kda/fla_cp.py      +105   CP: prefix-scan context, conv halo, fragment entry (new)

### DIFF
--- a/attn_gym/linear/kda/api.py
+++ b/attn_gym/linear/kda/api.py
@@ -27,6 +27,7 @@ from attn_gym.linear._delta_rule.validation import (
     validate_paged_state,
 )
 from attn_gym.linear.kda.constants import LOG2_E
+from attn_gym.linear.kda.impl.fla import chunk_forward as _fla_chunk_forward
 from attn_gym.linear.kda.impl.fused import chunk_forward as _fused_chunk_forward
 from attn_gym.linear.kda.impl.fused import paged_chunk_forward as _fused_paged_chunk_forward
 from attn_gym.linear.kda.impl.mega import chunk_forward as _mega_chunk_forward
@@ -131,6 +132,18 @@ def chunk_kda(
     )
     scale = resolve_scale(scale, q.shape[-1])
     if selected_impl is Impl.FUSED:
+        if backend == "fla":
+            return _fla_chunk_forward(
+                q,
+                k,
+                v,
+                gate,
+                beta,
+                initial_state,
+                cu_seqlens=cu_seqlens,
+                scale=scale,
+                output_final_state=output_final_state,
+            )
         if backend == "mega":
             return _mega_chunk_forward(
                 q,
--- a/attn_gym/linear/kda/impl/fla.py
+++ b/attn_gym/linear/kda/impl/fla.py
@@ -1,0 +1,66 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""KDA over flash-linear-attention's triton kernels.
+
+The fla kernels are triton, so this backend covers CUDA architectures the
+cute kernels do not. ``fla`` is an optional dependency; importing this module
+without it raises with an install pointer.
+
+The adapter feeds fla the same contract ``chunk_kda`` exposes: q/k already
+l2-normalized, ``gate`` the bound log-decay, ``beta`` already sigmoided --
+so every ``use_*_in_kernel`` switch stays off.
+"""
+
+import torch
+
+
+def _import_fla_chunk_kda():
+    try:
+        from fla.ops.kda import chunk_kda as fla_chunk_kda
+    except ImportError as err:
+        raise ImportError(
+            "kernel_options={'backend': 'fla'} needs flash-linear-attention: "
+            "pip install fla-core. It is an optional dependency, like the "
+            "cute kernels' cutlass."
+        ) from err
+    return fla_chunk_kda
+
+
+def chunk_forward(
+    q: torch.Tensor,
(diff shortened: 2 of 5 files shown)


QIU023:
cc @drisspg

drisspg (changes requested):
Thanks for putting this together,  we should forsure dispatch to FLA but I don't want to make FLA a first-class Impl or add its context-parallel API just yet.

Instead of using impl lets use kernel options. My hope is that most users will not need to know about how we run the computation but for practicality they might. kernel_options is that grab bag. Right now it supports backend: "mega" and "fused"(default).

If you look at our decode path we use a mix of cutedsl and triton but no tma so its pretty wide covering. I think this option is really about covering more than just blackwell. 

In the limit -> I think we will want: just fused and mega (perhaps none)

and fused will expand to cover the multi arch support through some combo of cutedsl and triton. 


So concretely; lets add "fla". Lets not add the CP in this pr

QIU023:
Hi Driss @drisspg, thanks for your detailed idea here! Have made the change followed above comments, `Impl.FLA` is deleted and `kernel_options` added `fla` along side `fused` and `mega`, and deleted the KCP from `fla` in this code base at all. LMK if this looks good to you

please keep me posted once you have follow plan about having KCP (Attention Gym's fused/triton version) somewhere so the https://github.com/pytorch/torchtitan/pull/4313 K3 CP PR can be rebased and rerun with that new KCP (on non-SM100/103).

drisspg:
hey hey @QIU023 sorry for the delay; I did a bunch of tweaking and I think we acutally no longer need this we should have full support for both gdn and kda on hopper (I think ampere) but need to find a machine and verify

QIU023:
NP! Verifying on my side and will cancel PR once can already unblock on latest HEAD KDA from your impl

QIU023:
Thanks for updating Driss! Verified on my fork that now latest KDA in attention gym with your #442 fix helped to normally run the portable kernel on RTX 5060Ti (SM120) that it's working, and totally agree that there is no need to put fla into this lib since you have written custom kda kernels for multiple architectures, cancelled this PR and waiting for your portable KDA kernel and its KCP to be applied to run together in https://github.com/pytorch/torchtitan/pull/4313 for K3 CP

QIU023:
close due to not needed anymore

drisspg:
@QIU023 Hey thanks again for putting up the pr and working on this! The portable kernels kda kernels and CP summary kernels should have landed by now im planning on tagging this morning

QIU023:
Sure thanks! Is it possible to expose these for CP  `context_parallel_kda()` and `PackedShard + `partition_packed_sequences`  to public callable APIs for  `attn_gym.linear.kda` so they can be called from titan?

drisspg:
yeah Im going to do a pass now on what thing is want to expose; I don't think our API will be as nice and 1 stop shop as it is for FLA since If people come up with cool new things I do don't want to have to add support for every new  thing but all the glue kernels will be public 

give me a sec

PR #449, opened 2026-09-02
Add an indexer kernel for DSA/CSA indexing

floatingtrees:
Right now, expressing this kernel (matmul + reduce over heads + topk) in pytorch materializes a `(batch, num_heads, seq, seq)` tensor in memory; at medium sequence lengths (4096), this tensor consumes 64 GB of memory and the indexing step consumes a similar amount of time as the core attention layer. This PR adds an API, reference, and CuTeDSL kernel for the indexer. 

The CuteDSL kernel runs around 5x faster than the default (175 vs 850 TFLOPs), and can also handle up to 1M without OOMing. Pytorch OOMs at `sequence_length>8192`

CuTeDSL Kernel Summary:

Each CTA is responsible for computing a `[batch_idx, 2 * query_index: 2 * query_index + 1, :]` chunk of the output. On a high level, it roughly performs the following steps.
1. Load 2 query tiles into SMEM; each query tile is in shape `(num_heads, head_dim)`. Also load key tiles into SMEM as well; key tiles are in shape `(sequence_length_chunk_size, head_dim)` 
2. Initialize `topk_buffer` and `staging_buffer`. 
3. While (not all key tiles computed yet)
a. Perform QK multiplication
b. Reduce across heads
c. Foreach `qk[i, j]` in tile, if `qk[i, j].score > min(topk_buffer)`, `staging_buffer.add(qk[i, j])`
d. If `staging_buffer.is_full`, `topk_buffer = sort(cat(staging_buffer, topk_buffer))[:topk_buffer.length]`, `staging_buffer.empty`
4. Write results back to global memory. 

The kernel uses 2 query tiles instead of 1 to improve tensor core utilization / arithmetic intensity. 

Limitations: Kernel doesn't work well for topk > 512. `torch.compile(fullgraph=true)` is also unsupported.


attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,149 @@
...
+        dots[b, t, h, s] = q[b, t, h, :] · k[b, s, :]
+        score[b, t, s]   = sum_h(w[b, t, h] * relu(dots[b, t, h, s]))
+                           / sqrt(H * D)
+        output[b, t, :]  = topk(score[b, t, :]).indices
+
+    Args:
+        q: Query tensor, [B, T, H, D].
+
+        k: Key candidate pool shared across heads, [B, S, D].
+            S may differ from T (nonsquare) unless causal=True.
```
eqy: this doesn't agree with the validation here right https://github.com/floatingtrees/attention-gym/blob/5a40759bf5caf042bfb96b13253c20e70ac829cb/attn_gym/sparse/indexer/api.py#L77-L81

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1668 @@
...
+        raise ValueError(
+            "prefill requires equal query/key lengths and matching B,D: "
+            f"q={tuple(q.shape)}, k={tuple(k.shape)}"
+        )
+    if tuple(weights.shape) != (batch, queries, heads):
+        raise ValueError(
+            f"weights must have shape {(batch, queries, heads)}, got {tuple(weights.shape)}"
+        )
+    if batch <= 0 or batch % 2:
+        raise ValueError(f"batch must be a positive multiple of 2, got {batch}")
```
eqy: this prevents the kernel from being used for decode? is this a real constraint?

attn_gym/sparse/indexer/impl/cute/impl.py
```diff
@@ -0,0 +1,1668 @@
...
+
+    Inputs are contiguous BF16/FP16 tensors with layouts q[B,T,H,D],
+    k[B,T,D], and weights[B,T,H].  The INT32 result has shape
+    [B,T,topk].  Invalid contracts, compilation errors, shared-memory
+    exhaustion, and launch errors propagate; there is no fallback.
+    """
+    if not isinstance(causal, bool):
+        raise TypeError(f"causal must be a bool, got {type(causal).__name__}")
+    _validate(q, k, weights, topk)
+    if topk == 0:
```
eqy, line 1576: This isn't reachable right due to https://github.com/floatingtrees/attention-gym/blob/5a40759bf5caf042bfb96b13253c20e70ac829cb/attn_gym/sparse/indexer/api.py#L123-L149

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,149 @@
...
+
+            return reference.index(q, k, weights, topk, causal)
+        case "triton":
+            raise NotImplementedError("Triton backend is not implemented yet.")
+        case "cute":
+            from .impl import cute as cute_backend
+
+            return cute_backend.index(q, k, weights, topk, causal)
+        case _:
+            raise NotImplementedError(f"Backend {backend!r} is not supported.")
```
eqy: codex comment is that these validations get bypassed if we early exit from topk=0

eqy (commented):
some superficial codex comments

attn_gym/sparse/indexer/impl/cute/impl.py
floatingtrees: Thanks for catching that, I'll fix it.

attn_gym/sparse/indexer/api.py
floatingtrees: This part is a bit outdated; the kernel supports nonsquare attention matrices, but I haven't decided how to add support for causality yet. I'll fix this comment.

attn_gym/sparse/indexer/api.py
floatingtrees: Ok, I'll move topk down.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: The kernel isn't decode optimized, since each CTA performs `Q_i * K_j` for all `j`. If we have batch size = 1, the kernel would be very slow. I think it's a leftover from some pipelining experiments; I'll remove it if it's not a real constraint.

attn_gym/sparse/indexer/impl/cute/impl.py
floatingtrees: I ended up removing the early topk return statement instead, so it still passes through all the checks as well.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1453 @@
...
+    batch, queries, heads, head_dim = q.shape
+    capability = torch.cuda.get_device_capability(q.device)
+    compile_key = (q.dtype, batch, queries, heads, head_dim, topk, causal, capability)
+    score_scale = 1.0 / math.sqrt(heads * head_dim)
+    q_c = from_dlpack(q, assumed_align=_ALIGNMENT)
+    k_c = from_dlpack(k, assumed_align=_ALIGNMENT)
+    weights_c = from_dlpack(weights, assumed_align=_ALIGNMENT)
+    output_c = from_dlpack(output, assumed_align=_ALIGNMENT)
+    compiled = _compile_cache.get(compile_key)
+    if compiled is None:
```
drisspg: Could we use the shared `jit_cache` + `compile_tvm_ffi` path with fake CuTe signatures, like the linear-attention kernels? This adds a separate cache and converts all four real tensors through DLPack on every call, even on cache hits. Keep real tensors at launch, and put the static specialization/fake ABI in the cached compiler.

test/test_indexer_cute.py
```diff
@@ -0,0 +1,308 @@
...
+    w = torch.randn(batch, queries, heads, device=device, dtype=dtype)
+
+    def run(q, k, w):
+        return index(q, k, w, topk, causal=causal, backend="cute")
+
+    eager_out = run(q, k, w)
+    assert eager_out.shape == (batch, queries, topk)
+    assert eager_out.dtype == torch.int32
+
+    compiled_run = torch.compile(run)
```
drisspg: Could we put the CuTe launcher behind a private `torch.library.define`/`impl` op with a fake implementation, then test the public API with `fullgraph=True` and `opcheck`, as we do for the linear-attention ops? Allowing graph breaks here only proves eager fallback runs; the `data_ptr()` check belongs inside the opaque launcher rather than being a reason to drop fullgraph support.

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,147 @@
...
+        backend: One of "eager", "triton", or "cute".
+
+        mode: Currently only prefill is supported; auto defaults to prefill.
+
+    Returns:
+        [B, T, topk] INT32 tensor of selected candidate indices. 
+        Not guaranteed to be sorted
+    """
+    if not torch.compiler.is_compiling():
+        _validate_inputs(q, k, weights, topk, causal)
```
drisspg: Could we keep the shape/dtype/device checks active under compilation? Skipping the whole validator changes the public contract: the eager backend's einsum accepts a broadcast batch dimension in `k`, and the weights expression accepts broadcast weights, while these inputs are rejected outside `torch.compile`. These metadata checks are traceable; only backend checks that inspect storage need an opaque boundary.

attn_gym/sparse/indexer/impl/cute/impl.py
```diff
@@ -0,0 +1,1453 @@
...
+    """
+    _validate(q, k, weights, topk)
+    if topk == 0:
+        return torch.empty((*q.shape[:2], 0), dtype=torch.int32, device=q.device)
+
+    output = torch.empty((*q.shape[:2], topk), dtype=torch.int32, device=q.device)
+    batch, queries, heads, head_dim = q.shape
+    capability = torch.cuda.get_device_capability(q.device)
+    compile_key = (q.dtype, batch, queries, heads, head_dim, topk, causal, capability)
+    score_scale = 1.0 / math.sqrt(heads * head_dim)
```
drisspg, line 1581: no from dl_pack point claude at my helpers for using make fake compact tensors

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1453 @@
...
+            k_c,
+            weights_c,
+            output_c,
+            score_scale,
+            topk,
+            _selection_sort_span(topk),
+            causal,
+            cute.runtime.make_fake_stream(),
+        )
+        _compile_cache[compile_key] = compiled
```
drisspg: also check otu how I setup the stream

test/test_indexer_cute.py
```diff
@@ -0,0 +1,308 @@
...
+    # FP64 scores at the kernel-selected positions
+    kernel_scores_64 = scores_64.gather(-1, safe_indices)  # [B, T, topk]
+
+    # Tolerance from the reduction chain: dot(D) then sum(H)
+    accumulation_eps = (
+        (math.sqrt(head_dim) + math.sqrt(heads)) * torch.finfo(torch.float32).eps
+    )
+    scale = torch.maximum(boundary.abs(), torch.ones_like(boundary))
+    tolerance = accumulation_eps * scale  # [B, T]
+
```
drisspg, line 265: Could we assert causal bounds separately from the score comparison? For rows with fewer than K causal keys, `boundary` is `-inf`, so selecting a future key also gives `-inf` and passes this check. Row 0 returning `[1, -1, ...]` passes the count and uniqueness checks too. Assert every valid index is <= its query position and every invalid slot is exactly -1.

test/test_indexer_cute.py
```diff
@@ -0,0 +1,308 @@
...
+        (2, 128, 64, 16, 32),
+        (2, 128, 64, 32, 32),
+        (2, 128, 64, 48, 32),
+        (2, 128, 64, 64, 32),
+        (2, 128, 64, 96, 32),
+        (2, 128, 64, 128, 0),
+        (2, 128, 64, 128, 1),
+        (2, 128, 64, 128, 127),
+        (2, 256, 64, 128, 129),
+        (2, 512, 64, 128, 512),
```
drisspg, line 168: Could we add a deterministic long-sequence case, e.g. S=1024/K=128 with better candidates in later tiles? The small-K cases stop at 256 candidates, and the S=512 case keeps everything. None exercises more input after the first selection-buffer drain, so cutoff updates, refill after pruning, and repeated mailbox reuse aren't covered by the current tests.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1453 @@
...
+_HEAD_DIM_GRANULARITY = 16
+_MMA_INSTRUCTION = (128, 64, 16)
+_MMA_TILE = (_TILE_CANDIDATES, _TILE_HEADS, _TILE_D)
+_QUERIES_PER_CTA = 2
+_SELECTION_WARPS = 4
+_SELECTION_THREADS = _SELECTION_WARPS * 32
+_CONSUMER_WARPS = _QUERIES_PER_CTA * _SELECTION_WARPS
+_K_STAGES = 2
+_Q_STAGES = 4
+_MAILBOX_STAGES = 4
```
drisspg: Could we follow the class-based kernel + typed configuration structure used in `linear/short_conv/cute.py` here? Stage counts, query tiling, and warp-role assignments are currently module-global policy shared implicitly by the storage type and all the helpers. I'd keep that schedule owned by the indexer kernel/config, with named warp roles, rather than adding another implementation convention.

attn_gym/sparse/indexer/impl/reference.py
```diff
@@ -0,0 +1,58 @@
...
+    batch, queries, heads, head_dim = q.shape
+    candidates = k.shape[1]
+    if topk == 0:
+        return torch.empty((batch, queries, 0), dtype=torch.int32, device=q.device)
+    scale = 1.0 / math.sqrt(heads * head_dim)
+
+    # dots: [B, T, H, S]
+    dots = torch.einsum("bthd,bsd->bths", q, k)
+    # score: [B, T, S]
+    scores = (torch.relu(dots) * weights.unsqueeze(-1)).sum(dim=2) * scale
```
drisspg, line 51: Could we define one accumulation contract across backends? This materializes the dots and weighted products in the input dtype, while CuTe keeps FP32 scores. With FP16, even finite inputs can overflow the dot and mixed-sign weights then produce NaNs before scaling. The `matches_eager` test misses this because it uses promoted FP32 inputs rather than calling this backend. I'd promote low-precision inputs for reference scoring and retain FP64 when supplied.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1453 @@
...
+                        scope="cta",
+                    )
+                )
+            warp_base = Int32(cute.arch.shuffle_sync(warp_base, 0))
+            if accept:
+                selection_keys[query_base + Int32(run_size) + warp_base + lane_rank] = candidate_key
+
+            selection_barrier.arrive_and_wait()
+            occupied = Int32(buffer_counts[query_slot])
+            if occupied > Int32(0) and occupied >= Int32(drain_threshold):
```
drisspg: Could we synchronize after every thread snapshots `buffer_counts`, before branching? The barrier above orders the previous appends, but a fast warp can read 224, skip the drain, and append 32 candidates from the next tile before a slower warp reads the counter. That warp sees 256 and enters the drain for the previous tile, so the group disagrees about which barriers/sort to execute. This is a blocking race by inspection, not GPU-reproduced. Please cover partial acceptance followed by another tile, e.g. S=512/K=128 with tile acceptance counts 128, 96, 128 after seeding.

attn_gym/sparse/indexer/impl/reference.py
```diff
@@ -0,0 +1,58 @@
+"""Pure-PyTorch reference indexer implementation."""
+
+import math
+
+import torch
+from torch import Tensor
+
+
+def index(
```
drisspg, line 9: OOC would any changes be needed for qwnens sparse attention indexer?

test/test_indexer_cute.py
floatingtrees: Ok, thanks for catching this. Do you think we need a fully deterministic long sequence test case, or is a sufficiently long sequence test (like S=4096, K=128) enough?

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: That makes sense, I fixed the issue

attn_gym/sparse/indexer/impl/reference.py
floatingtrees: Right now, the main issue that that the kernel's MMA size uses `N=num_heads`, and qwen's indexer only uses 4 heads, so the MMA would be really inefficient. It should be possible to implement some packing scheme that computes the scores for multiple index positions in one MMA though. 

The reduction scheme over heads is different (Deepseek uses a relu and Qwen doesn't). I was thinking about letting users define a reduction over heads, but I'm not sure how much complexity that would add or if it would be worth it.

attn_gym/sparse/indexer/api.py
floatingtrees: Ok, that makes sense. I fixed the issue.

attn_gym/sparse/indexer/impl/cute/impl.py
floatingtrees: I switched over to using the shared cache, so this issue should be fixed now.

attn_gym/sparse/indexer/impl/reference.py
floatingtrees: Ok, that makes sense. I promoted lower precision dtypes inside of the function to fp32 and kept fp64 inputs in fp64.

test/test_indexer_cute.py
floatingtrees: Thanks for catching that, I added a causality check.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Thanks for pointing this out, your way seems a lot more convenient.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Ok, I added a new synchronization block that fixes this issue and an extra test case.

test/test_indexer_cute.py
floatingtrees: Ok, I fixed this, and `fullgraph=True` works now.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Ok, I refactored code and fixed this issue.

floatingtrees:
@drisspg Thanks for the comments! I think I addressed everything, let me know if you see anything else!

attn_gym/sparse/indexer/impl/reference.py
drisspg: That makes sense. For short, I added this registration mechanism for having users be able to register any pointwise operator, as long as they register a forward and a backward. Could be interesting here. We can do that in a follow-up.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+    agree on lives here; nothing about the schedule is a module-global
+    constant. ``IndexerPrefillKernel`` owns one instance and every helper
+    that needs schedule state reads it from ``self.config``.
+    """
+
+    tile_candidates: int = 128
+    tile_heads: int = 64
+    tile_d: int = 64
+    queries_per_cta: int = 2
+    selection_warps: int = 4
```
drisspg: Could we distinguish fixed kernel invariants from actual configuration choices? `queries_per_cta` looks configurable, but the kernel has exactly two query pipelines; the sorter still assumes 128 keys/four keys per lane, and the reducer assumes 64 heads per tile. I'd encode those fixed values on the kernel and retain only supported config choices. With four selection warps fixed, the role enum can also hold the actual warp boundaries instead of going through two dictionaries in `warp_range`.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+        mQ_hdtb: cute.Tensor,
+        mW_bth: cute.Tensor,
+        mOut_btk: cute.Tensor,
+        k_smem_layout: cute.ComposedLayout,
+        q_smem_layout: cute.ComposedLayout,
+        io_dtype: cutlass.Constexpr,
+        score_scale: Float32,
+        topk: cutlass.Constexpr,
+        sort_span: cutlass.Constexpr,
+        causal: cutlass.Constexpr,
```
drisspg: Could these methods read `topk`, `causal`, and `sort_span` from the owning instance instead of retaining the old free-function parameter chain? They are already class-owned specialization state, but `__call__` forwards them into `kernel`, which forwards them again and reconstructs `run_size` from `sort_span`. The class should remove that plumbing, not just wrap it.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+        raise ValueError(f"q, k, and weights must be {_ALIGNMENT}-byte aligned")
+    if torch.cuda.get_device_capability(q.device) != (10, 0):
+        raise RuntimeError("this tcgen05 kernel requires an SM100 GPU")
+
+
+@jit_cache
+def _compile_indexer(
+    dtype_name: str,
+    batch: int,
+    queries: int,
```
drisspg: Could B/T be symbolic in the fake ABI rather than compile-cache dimensions? The device code already derives the grid and loop bounds from tensor shapes; the stored batch/query counts are only used in the artifact name. This still recompiles for every sequence length despite the dynamic-fullgraph test. The affine-summary kernels show the symbolic-token pattern; please check reuse of one CuTe artifact across B/T changes, not just reuse of the Dynamo graph.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+kernel.  The launcher is guarded by an
+in-process and persistent TVM-FFI compile cache keyed on the static shape/dtype
+contract (dtype, batch, queries, heads, head_dim, topk, causal) and compile
+target. The public API invokes it through the private operator in
+``attn_gym.sparse.indexer.ops``; there is no fallback implementation.
+
+The launch schedule (stage counts, query/candidate tiling, warp-role
+assignment) is owned by ``IndexerConfig`` and ``IndexerWarpRole`` and threaded
+through ``IndexerPrefillKernel`` instance state, rather than being implicit
+module-global policy shared by free functions.
```
drisspg: nit: could we drop the repeated explanation of moving globals into a class? The module header, `IndexerConfig`, and `IndexerPrefillKernel` all narrate the refactor. Keep the useful pipeline/dataflow description and the fixed scheduling constraints; the ownership should be evident from the code.

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def index(
+    q: Tensor,
+    k: Tensor,
+    weights: Tensor,
+    topk: int,
+    causal: bool = False,
+    backend: str = "eager",
+    mode: str = "auto",
```
drisspg: Could we drop `mode` until there are genuinely different implementations to select? `auto` and `prefill` both do nothing, and `triton` is also advertised despite always raising. I'd expose only the implemented backend choices and make `causal`/`backend` keyword-only after `topk`, rather than reserving a larger public API for future paths.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+    sequence length or compile target selects a different specialization.
+    """
+    _validate(q, k, weights, topk)
+    if topk == 0:
+        return torch.empty((*q.shape[:2], 0), dtype=torch.int32, device=q.device)
+
+    output = torch.empty((*q.shape[:2], topk), dtype=torch.int32, device=q.device)
+    batch, queries, heads, head_dim = q.shape
+    score_scale = 1.0 / math.sqrt(heads * head_dim)
+    set_compile_target(detect_compile_target(q.device.index))
```
drisspg: Could we scope this target override and restore the previous target even if compilation fails? `set_compile_target` changes process-global state, so an ordinary indexer call silently replaces any explicit target and leaves that choice installed for subsequent unrelated kernels. Selecting metadata from `q.device` is right; leaking that selection beyond this operation isn't.

attn_gym/sparse/__init__.py
```diff
@@ -1,5 +1,6 @@
 """Sparse attention primitives."""
 
+from .indexer import index
```
drisspg: lets be more literal and use `indexer_topk`

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,140 @@
...
+        raise ValueError(f"topk must be in [0, {candidates}], got {topk}.")
+
+    # --- square constraint (nonsquare not yet supported) ---
+    if candidates != queries:
+        raise NotImplementedError(
+            f"Nonsquare inputs are not supported yet (T={queries}, S={candidates})."
+        )
+
+
+def index(
```
drisspg: yeah indexer_topk or lightning_indexer

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,140 @@
...
+        dots[b, t, h, s] = q[b, t, h, :] · k[b, s, :]
+        score[b, t, s]   = sum_h(w[b, t, h] * relu(dots[b, t, h, s]))
+                           / sqrt(H * D)
+        output[b, t, :]  = topk(score[b, t, :]).indices
+
+    Args:
+        q: Query tensor, [B, T, H, D].
+
+        k: Key candidate pool shared across heads, [B, S, D].
+            Nonsquare inputs (S != T) are not supported yet.
```
drisspg, line 103: shoudl we support GQA as well or just MQA?

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,140 @@
...
+    _validate_inputs(q, k, weights, topk, causal)
+
+    match mode:
+        case "auto":
+            pass  # auto currently dispatches to prefill
+        case "prefill":
+            pass
+        case _:
+            raise NotImplementedError(f"Mode {mode!r} is not supported.")
+
```
drisspg: yeah jsut dro pthese

attn_gym/sparse/indexer/api.py
```diff
@@ -0,0 +1,140 @@
...
+    match backend:
+        case "eager":
+            from .impl import reference
+
```
drisspg: backend = kernel backend and impl = reference vs fused;  lets just ahve an impl arg it defualts to fused and hten add a `kernel_options` dictionary and in teh future if osoene wnast triton or cute or what hav eoyu thatt we cant auto figure out they can specify see the KDA and gdn apis for example

attn_gym/sparse/indexer/impl/cute/prefill.py
drisspg: To clarify: derived properties are good here. `topk` and the input shapes are configurable, and `run_size`/`sort_span` should follow from them. The distinction is that two query pipelines, 128-candidate tiles, and four selection warps are assumptions baked into this implementation, not currently supported tuning choices. Could we encode those fixed invariants on the kernel and keep the config for genuinely supported choices?

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+        q_barriers: cute.struct.MemRange[Int64, config.q_stages * 2]
+        acc0_barriers: cute.struct.MemRange[Int64, config.acc_stages * 2]
+        acc1_barriers: cute.struct.MemRange[Int64, config.acc_stages * 2]
+        mailbox0_barriers: cute.struct.MemRange[Int64, config.mailbox_stages * 2]
+        mailbox1_barriers: cute.struct.MemRange[Int64, config.mailbox_stages * 2]
+        tmem_holding: Int32
+
+    return SharedStorage
+
+
```
drisspg: Could we define `SharedStorage` inside `IndexerPrefillKernel.__init__` and assign it to `self.SharedStorage` instead of using a module-level factory? The array extents can still come from the config, while keeping the shared-memory layout owned by the kernel class.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+    def threads(self) -> int:
+        """Total CTA thread count: every warp role, sized by ``config.selection_warps``.
+
+        ``IndexerWarpRole`` values are ordinal role markers, not literal warp
+        indices -- ``warp_range`` is the source of truth for actual warp
+        indices, since those scale with ``config.selection_warps``.
+        """
+        return self.warp_range(IndexerWarpRole.LOAD).stop * _WARP_SIZE
+
+    def warp_range(self, role: "IndexerWarpRole") -> range:
```
drisspg: For an existing example, see `WarpRole` in `attn_gym/linear/_delta_rule/cute/affine_summary_fwd.py:100` and its `CUDA_WARP_IDS`/`CTA_THREADS` definitions at lines 153–155. The enum holds actual warp boundaries, and ranges/thread counts are derived directly. `linear/kda/fwd/cute/chunk_kda_fwd_intra_engine.py:123–151` follows the same pattern. With this kernel's fixed schedule, boundaries 0/4/8/12/16/17/18 would replace these two dictionaries. If variable warp counts become supported, deriving ranges from a validated config would make sense instead.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+                query_base,
+                tidx,
+                query_slot,
+                selection_barrier,
+                run_size,
+                sort_span,
+            )
+
+    @cute.kernel
+    def kernel(
```
drisspg: Super nit; . I like having kernel towards the top because then you can read it and then jump and read it from top to bottom. It's a little bit easier.

attn_gym/sparse/indexer/impl/cute/prefill.py
```diff
@@ -0,0 +1,1696 @@
...
+
+    @property
+    def mma_tile(self) -> tuple[int, int, int]:
+        return (self.tile_candidates, self.tile_heads, self.tile_d)
+
+
+class IndexerWarpRole(IntEnum):
+    """Warp-role boundaries in the prefill kernel.
+
+    Each named role spans a range of warps rather than a single warp index;
```
drisspg: nit lets not call this prefill.py right maybe impl.py idk what we call the other oens but prefil seems really specific

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Ok, I hardcoded the kernel invariants and used the raw warp boundaries.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: That makes sense, I fixed it.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Ok, I fixed this

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Sorry about that, I deleted the useless comments.

attn_gym/sparse/indexer/api.py
floatingtrees: Ok, this issue is fixed.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: That makes sense, I updated the code to save the previous target and reload it after compilation finishes/crashes.

attn_gym/sparse/__init__.py
floatingtrees: Ok, I renamed it to `lightning_indexer` instead

attn_gym/sparse/indexer/api.py
floatingtrees: Renamed to `lightning_indexer`

attn_gym/sparse/indexer/api.py
floatingtrees: Ok, those are gone now

attn_gym/sparse/indexer/api.py
floatingtrees: I think users who want GQA can reshape their tensors to `(batch * kv_heads, total_heads // kv_heads, sequence, head_dim)`, assuming they want an output tensor in shape of `(batch, kv_heads, queries, topk)`, so GQA is decently supported already.

attn_gym/sparse/indexer/api.py
floatingtrees: Ok, I fixed this issue.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Ok, I moved it into `__init__`

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: Ok, this is fixed as well.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: I never that of that, it makes sense though. I moved it up.

attn_gym/sparse/indexer/impl/cute/prefill.py
floatingtrees: I renamed it to impl.py. Maybe it will have to be re-renamed if we ever get a decode kernel though.

floatingtrees:
@drisspg thanks for the feedback! I addressed everything, tests pass and linting is good too.

drisspg (approved):

PR #453, opened 2026-09-02
Add staged delta-rule primitives and a context-parallel recipe for KDA and GDN

drisspg:
### What this adds

Context parallelism for the delta-rule ops (KDA and GDN) as a small set of composable pieces instead of
one monolithic `chunk_kda_cp`. The idea: every delta-rule token step is affine in the recurrent state, so
any run of tokens collapses to one FP32 map `H_out = H_in @ A + B`, stored as `[HV, V+K, K] = [bias;
transition]` (one map per value head). Each rank summarizes the tokens it owns, ranks exchange those summaries, each rank folds
its predecessors' summaries into an entry state, and runs the ordinary local kernel. Backward is the
mirror image with a reverse summary `dH_in = dH_out @ R + C`. Everything here is built from that fact.

Follow-up to the discussion on #421: the library exposes capability, not a recipe that grows a flag per
partition or topology.

### Using it

**One call, the common case.** You decide which global token ranges each rank owns; the plan does the rest.

```python
from attn_gym.linear.context_parallel import ContextParallelPlan
from attn_gym.linear.kda import context_parallel_kda          # or attn_gym.linear.gdn.context_parallel_gdn

# Every rank builds the same table. Here: zig-zag over 2W equal blocks (what ring softmax CP uses).
blocks = 2 * world_size
size = total_tokens // blocks
token_ranges = [[(r * size, (r + 1) * size), ((blocks - 1 - r) * size, (blocks - r) * size)]
                for r in range(world_size)]

plan = ContextParallelPlan.from_token_ranges(global_cu_seqlens, token_ranges, rank)
token_ids = plan.global_token_ids(device)                     # slice global tensors into this rank's local ones
cu_seqlens = torch.tensor(plan.cu_seqlens, dtype=torch.int32, device=device)

output, final_state = context_parallel_kda(
    q, k, v, gate, beta, cu_seqlens=cu_seqlens, plan=plan, group=cp_group
)
true_final_states = final_state[list(plan.terminal)]         # only fragments that end their sequence
```

Contiguous shards are `[[(r * size, (r + 1) * size)] for r in range(world_size)]`; document-aligned or
uneven partitions are just other lists. The plan cuts ranges at sequence boundaries into `Fragment`s, one
local `cu_seqlens` segment each, so two pieces of the same document on one rank are simply two local
segments whose entry states the recipe supplies. The short-conv halo uses the same plan:
`context_parallel_conv_history(qkv, plan, group, kernel_size - 1)` returns `causal_conv1d`'s
`initial_state`.

**The primitives, when you want your own communication.** This is what the recipe is made of:

```python
from attn_gym.linear.kda import chunk_kda_prepare, chunk_kda_prepare_backward
from attn_gym.linear.state_summary import merge_state, compose_summaries, neutral_summary

# forward
prepared = chunk_kda_prepare(q, k, v, gate, beta, cu_seqlens=cu_seqlens)   # WY factors, computed once
summary = prepared.state_summary(start, stop)        # [bias; transition] of one local sequence, FP32
# ... move summaries however you like ...
entry = merge_state(zero_state, predecessor_summary)  # fold predecessors left to right from zero
output, final_state = prepared.run(entry_states, output_final_state=True)

# backward (inside your autograd.Function; prepared.saved is the NamedTuple to save_for_backward)
grads = chunk_kda_prepare_backward(saved, d_output, entry_states)
grad_summary = grads.state_grad_summary(start, stop)  # reverse map [C; R]
# ... exchange, fold successors from the far end ...
dq, dk, dv, dgate, dbeta, _ = grads.run(exit_cotangents)
```

`start`/`stop` are host ints naming exactly one local `cu_seqlens` segment, so nothing syncs and CUDA
Graph capture works. `compose_summaries(first, then)` combines two maps before any state is known (for
scans). `neutral_summary` is the identity, the right filler for a collective slot nobody reads.

### What is supported

- KDA and GDN (including GQA, `HK < HV`), fp16/bf16 Q/K/V, packed `cu_seqlens`, arbitrary fragment
  sizes (partial chunks are handled), any number of fragments per rank (unequal counts are padded in the
  all-gather).
- Exact final states for every fragment; `plan.terminal` tells you which ones are real sequence ends.
  The loss may attach to any of them.
- Full backward with all five input gradients. Against an FP32 reference the sharded gradients have the
  same error as the unsharded fused op (checked to 2-3 digits on 2 GPUs for both ops, both layouts).
- CUDA Graph capture of the whole forward + collectives + backward, as long as the plan is fixed.
- Hopper and datacenter Blackwell (summary kernels: Triton on Hopper, native UMMA on SM100).

### What it assumes

- Every sequence starts from the zero state; `initial_state` is not accepted under the recipe (same
  stance as FLA). Use the primitives if you need something else.
- `token_ranges` must tile `[0, total_tokens)` exactly once, every rank must own at least one range, and
  all ranks build the same table. `check_group` rejects a plan whose world size, rank, or local token
  count disagrees with the group and input.
- The device `cu_seqlens` you pass must be `plan.cu_seqlens`; this is not verifiable without a sync, so
  it is a contract.
- A captured CUDA Graph is valid for one plan; changing the partition means recapturing.
- B = 1 (pack sequences), matching fp16/bf16 Q/K/V, and the fused kernels' `K = V = 128`.

### How to extend

- **Another partition:** write a different `token_ranges` list. Nothing else changes.
- **Another topology** (P2P pipeline, recursive-doubling scan over `compose_summaries`, DTensor,
  comm/compute overlap): copy `_ContextParallelChunk` in `attn_gym/linear/context_parallel.py` and swap
  the collective. The primitives and plans are unchanged; the compose helpers are pure functions you can
  reuse.
- **Another delta-rule op:** provide `prepare` / `prepare_backward` with the same handle protocol and pass
  them as a `StagedOp`. GDN is the worked example: `gdn/impl/chunk.py` is split into the same
  prepare/finish seams (behavior-preserving; public `chunk_gdn_bwd` composes them), `gdn/stages.py` wraps
  them, and `gdn/context_parallel.py` is 45 lines of binding.
- **Mega backend:** not wired. Mega keeps the WY factors on chip so there is no `prepare` output for the
  summary kernel; a Mega summary mode would slot in as another `StagedOp`.

### Room to grow (deliberately not in this PR)

```text
comm/compute overlap    the shipped recipe blocks on one all-gather per direction; the primitives let a
                        caller launch state_summary, start the collective, run other layer work, then run
scan instead of gather  the gathered buffer is W x slots x summary; recursive doubling over
                        compose_summaries would be O(log W) exchanges of one summary each
fully fused forwards    Mega keeps WY factors on chip, so there is no prepare output to summarize; a Mega
                        summary mode (augmented [0 | I] state) is the follow-up
```

### Review order

`state_summary.py` -> `kda/stages.py` -> `gdn/impl/chunk.py` (pure split) -> `gdn/stages.py` ->
`context_parallel.py` -> the two thin bindings -> example -> tests. Two fixes surfaced by review along the
way: the staged path skipped `normalize_tma_rows`, so strided Q/K/V that `chunk_kda` accepts failed in
TVM-FFI, and `beta` slices need `normalize_compact_tensor` rather than `.contiguous()`.

## Test Plan

```bash
pytest -n 6 test/test_delta_rule_stages.py test/test_kda_context_parallel.py test/test_namespaces.py \
  test/test_kda_training_example.py test/test_delta_affine_summary.py test/test_short_conv_cute.py \
  test/linear/test_gdn_chunk_fused.py test/linear/test_gdn_fused.py test/linear/test_gdn.py \
  test/test_mega_optional_import.py   # 323 passed, 6 skipped (GB200)
CUDA_VISIBLE_DEVICES=0,1 pytest test/test_context_parallel_distributed.py   # 4 passed, 2 x GB200
torchrun --standalone --nproc-per-node=2 examples/kda_context_parallel.py
torchrun --standalone --nproc-per-node=2 examples/kda_context_parallel.py --partition zigzag
torchrun --standalone --nproc-per-node=2 examples/kda_context_parallel.py --partition zigzag --compute-dtype float16
torchrun --standalone --nproc-per-node=2 examples/kda_context_parallel.py --cuda-graph
torchrun --standalone --nproc-per-node=2 examples/kda_context_parallel.py --partition zigzag --cuda-graph
ruff check . && ruff format --check attn_gym examples test && mkdocs build
```

`test_delta_rule_stages.py` simulates the full recipe single-process for kda / gdn / gdn-gqa across
contiguous-2, zigzag-2, zigzag-3, and an uneven reordered table in bf16 and fp16, and asserts sharded
gradient error against an FP32 oracle is no worse than the unsharded fused op's.
`test_context_parallel_distributed.py` (skipped below two GPUs) spawns two NCCL ranks and runs the public
`context_parallel_kda` / `context_parallel_gdn` themselves on a zig-zag plan, with and without a
final-state loss term, so the autograd Function, the all-gather, and the `d_final_state=None` backward
path are exercised end to end against the unsharded op. `compose_entry_states` / `compose_exit_cotangents`
fold shared chain prefixes once (`_fold_chains`, keyed by prefix), which keeps a rank that owns many
pieces of one sequence at O(pieces) merges; a test pins 7 merges for 8 pieces instead of 22.

### DIFF
--- a/attn_gym/linear/context_parallel.py
+++ b/attn_gym/linear/context_parallel.py
@@ -1,0 +1,455 @@
+# Copyright (c) 2025 Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Context parallelism for delta-rule attention: ownership plans, state routing, one recipe.
+
+Data model. The global packed stream is a list of sequences. Each rank owns any set of global token
+ranges and lays them out back to back as its local packed input. The plan cuts those ranges at
+sequence boundaries into :class:`Fragment` pieces, one local ``cu_seqlens`` segment each, so
+contiguous shards, zig-zag load balancing, and document-aligned partitions are just different range
+lists. Because the recurrence is affine (see ``attn_gym.linear.state_summary``), a fragment's entry
+state is its predecessors' summaries folded from zero and its exit cotangent is its successors'
+reverse summaries folded from zero.
+
+Communication model. Every rank fills ``plan.slots`` summary slots in local order (identity where
+nobody downstream needs one) and all-gathers them once per direction, so every
+``gathered[rank][slot]`` index is a host-static integer and CUDA Graph capture works as long as the
+plan does not change. The short-convolution halo reuses the plan: each fragment's history is the
+tail of its predecessor fragments.
+
+Ops. ``context_parallel_chunk`` is generic over the delta-rule variant through a :class:`StagedOp`
+pair of ``prepare`` / ``prepare_backward`` callables whose handles expose ``state_summary``,
+``run``, and ``saved`` (``attn_gym.linear.kda.stages``, ``attn_gym.linear.gdn.stages``). It is one
+recipe, not an extension point: for a point-to-point pipeline, a recursive-doubling scan over
+``compose_summaries``, DTensor, or communication overlap, copy the autograd function and swap the
+collective.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from itertools import accumulate, pairwise
+from typing import NamedTuple, Protocol
+
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+from torch.distributed._functional_collectives import all_gather_single
+
+from attn_gym.linear.kda.utils import profiler_range
+from attn_gym.linear.state_summary import merge_state, neutral_summary
+
+
+@dataclass(frozen=True)
+class Fragment:
+    """A contiguous global token range ``[start, stop)`` lying inside one logical sequence."""
+
+    sequence: int
+    start: int
+    stop: int
+
+    @property
+    def length(self) -> int:
+        return self.stop - self.start
+
+
+# ``(rank, slot)`` address of a fragment in every rank's gathered buffer.
+Slot = tuple[int, int]
+
+
+def _cut_at_sequences(
+    global_cu_seqlens: Sequence[int], token_ranges: Sequence[tuple[int, int]]
+) -> tuple[Fragment, ...]:
+    fragments = []
+    for range_start, range_stop in token_ranges:
+        for sequence, (seq_start, seq_stop) in enumerate(pairwise(global_cu_seqlens)):
+            start, stop = max(range_start, seq_start), min(range_stop, seq_stop)
+            if start < stop:
+                fragments.append(Fragment(sequence, start, stop))
+    return tuple(fragments)
+
+
+def _validate_tiling(
+    global_cu_seqlens: Sequence[int], token_ranges: Sequence[Sequence[tuple[int, int]]]
+) -> None:
+    """Require the ranks' token ranges to tile ``[0, total_tokens)`` exactly once."""
+    if len(global_cu_seqlens) < 2 or global_cu_seqlens[0] != 0:
+        raise ValueError("global_cu_seqlens must start at zero and contain at least one sequence")
+    if any(end < start for start, end in pairwise(global_cu_seqlens)):
+        raise ValueError("global_cu_seqlens must be nondecreasing")
+    if not token_ranges or any(not ranges for ranges in token_ranges):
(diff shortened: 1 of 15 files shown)


drisspg:

PR #455, opened 2026-09-02
Run the Mega backend under context parallelism

drisspg:
## Agent note

Stacked on #453. `chunk_kda_prepare(..., kernel_options={"backend": "mega"})` and
`context_parallel_kda(..., kernel_options=...)` now run each rank's local pass on the Mega kernel from the
composed entry state, which is exactly how Mega already runs with an `initial_state`.

The split follows what Mega can and cannot do. Job B, running a fragment from its entry state to output
and exit state, is `chunk_mega_packed_fwd_with_state_op` unchanged; nothing is materialized. Job A,
producing a fragment's `[bias; transition]` summary, needs the WY factors that Mega keeps on chip, so
`ChunkKDAMegaPrepared.state_summary` factors just the requested sequence (a whole local segment, so its
chunk grid is unchanged when treated as a stream of its own); ranks whose fragments all end their sequence
never pay for it, and packed streams pay only for the fragments that continue on another rank. For packed-document workloads the summarized fragment
is a few hundred tokens and the extra pass is negligible; for one long sequence per rank it is one intra
pass plus the summary kernel over the shard, on top of Mega's on-chip run. Backward is Mega's existing
with-state route: `ChunkKDASaved.aqk/akk` may now be `None`, and `chunk_kda_prepare_backward` rebuilds
them with `chunk_kda_fwd_factors` the way `_chunk_kda_bwd_shared` does, then reuses the fused staged
backward and the reverse summary kernel unchanged. `split_backward` is rejected because staged runs always
carry a state.

Also fixes the public Mega wrapper's dense no-state path, which rejected `T % 64 != 0` with "pass
cu_seqlens for tails". Inputs are not padded to 64 in practice; a partial tail now takes the packed path
with synthesized `[0, T]` boundaries, exactly as the fused backend and Mega's own with-state route already
did. The test that pinned the rejection now asserts the tail result equals the explicit-boundary result.

Against an FP32 reference on two GB200s, sharded Mega gradients match unsharded Mega error to 2-3 digits
for contiguous and zig-zag plans. The Mega summary mode (augmented `[0 | I]` state on chip) remains the
follow-up that would remove job A's HBM pass; it would replace `state_summary` behind the same handle.

## Performance

Mega under CP is faster exactly where Mega is faster on one GPU: many sequences or many heads. Its
persistent kernel has one work item per (sequence, head), so one long sequence at H=4 leaves most SMs idle
regardless of CP, while packed documents fill the machine. Two GB200s, bf16, H=4, K=V=128, one
`context_parallel_kda` forward+backward step, CUDA-event timing, max over ranks, best of 3 interleaved
rounds, same seeded tensors and plan for both backends (`agent_space/bench_cp_mega.py`):

| workload (global)            | plan       | fused CP | Mega CP  | fused / Mega |
|------------------------------|------------|---------:|---------:|-------------:|
| 1 sequence, 131k tokens      | contiguous |  2.78 ms |  5.19 ms | 0.54x        |
| 1 sequence, 524k tokens      | contiguous | 10.90 ms | 20.46 ms | 0.53x        |
| 1 sequence, 1M tokens        | contiguous | 21.69 ms | 40.84 ms | 0.53x        |
| 1 sequence, 1M tokens        | zig-zag    | 18.30 ms | 26.76 ms | 0.68x        |
| 1024 x 1k-token docs, 1M     | contiguous |  6.80 ms |  2.32 ms | 2.94x        |

Single-GPU reference on the same per-rank shapes (`chunk_kda`, no CP): 524k x 1 sequence fused 11.8 ms
vs Mega 28.0 ms (0.42x); 524k as 512 docs fused 5.9 ms vs Mega 1.3 ms (4.4x); at H=32, 131k x 1
sequence Mega is 1.5x faster and 128 docs 4.8x. CP adds about 1 ms per step for either backend (summary
kernels on the boundary fragments, two all-gathers, composes), which is why the packed-docs ratio is 2.9x
under CP versus 4.4x alone. For one long sequence the dominant CP cost is the forward/reverse summary
kernel over the whole shard (roughly equal to the local forward+backward itself); that is the serial
per-chunk chain in K1/K2, not the backend, and zig-zag halves it by halving fragment length.

## Test Plan

```bash
# .venv-mega: -e '.[mega,dev]' (CuTeDSL 4.7.1) on SM100; default env skips the kda-mega cases
pytest -n 4 test/test_delta_rule_stages.py                     # 50 passed (mega env)
pytest -n 6 test/test_delta_rule_stages.py test/test_kda_context_parallel.py \
  test/test_namespaces.py test/test_kda_training_example.py     # 69 passed, 12 skipped (default env)
torchrun --standalone --nproc-per-node=2 agent_space/cp_gdn_smoke.py kda contiguous notune mega
torchrun --standalone --nproc-per-node=2 agent_space/cp_gdn_smoke.py kda zigzag notune mega
ruff check . && mkdocs build
```

### DIFF
--- a/attn_gym/linear/kda/context_parallel.py
+++ b/attn_gym/linear/kda/context_parallel.py
@@ -8,6 +8,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from functools import partial
 
 import torch
@@ -30,14 +31,17 @@ def context_parallel_kda(
     scale: float | None = None,
     autotune: bool = True,
     fastmath: bool = False,
+    kernel_options: Mapping[str, object] | None = None,
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    """Run fused ``chunk_kda`` over this rank's fragments with state exchanged by all-gather.
+    """Run ``chunk_kda`` over this rank's fragments with state exchanged by all-gather.
 
     See ``attn_gym.linear.context_parallel.context_parallel_chunk`` for the argument contract;
-    ``scale``, ``autotune``, and ``fastmath`` follow ``chunk_kda``.
+    ``scale``, ``autotune``, ``fastmath``, and ``kernel_options`` follow ``chunk_kda``. With
+    ``kernel_options={"backend": "mega"}`` the local pass runs on Mega and fused factors are
+    computed only for fragments that must send a summary.
     """
     stages = StagedOp(
-        partial(chunk_kda_prepare, scale=scale, autotune=autotune),
+        partial(chunk_kda_prepare, scale=scale, autotune=autotune, kernel_options=kernel_options),
         partial(chunk_kda_prepare_backward, scale=scale, autotune=autotune, fastmath=fastmath),
     )
     return context_parallel_chunk(
--- a/attn_gym/linear/kda/impl/mega.py
+++ b/attn_gym/linear/kda/impl/mega.py
@@ -321,10 +321,6 @@ def validate_mega_constraints(
             raise TypeError("Mega beta requires a contiguous, element-aligned inner mode")
         if cu_seqlens is not None and cu_seqlens.data_ptr() % 8:
             raise TypeError("Mega cu_seqlens must be 8-byte aligned")
-    if cu_seqlens is None and q.shape[1] % _CHUNK_SIZE:
-        raise ValueError(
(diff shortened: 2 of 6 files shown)


drisspg:

PR #479, opened 2026-09-04
Route context parallelism through device tensors so one graph replays across layouts

drisspg:
Stacked PRs:
 * __->__#479
 * #478
 * #477
 * #474
 * #473
 * #472
 * #471
 * #470

--- --- ---

Route context parallelism through device tensors so one graph replays across layouts

The plan's gather slot becomes one per fragment instead of one per subsequence: forward, the
summary of the fragment's last subsequence (only it can continue on another fragment); backward,
the reverse map of its first. That is strictly less communication than before, and it makes every
shape a function of the fragment table alone: slots per rank, and the longest chain, which is the
number of fragments in the table. `Fragment` joins `Subsequence` in the plan; `summary_slots`,
`grad_summary_slots`, and the short-convolution halo follow the new slot model.

`plan.routing(device, max_subsequences=N)` materializes a layout as fixed-shape device tensors
(`ContextParallelRouting`): the span's `cu_seqlens` padded with empty segments to `N`, one
`[start, stop)` per local fragment for `state_summaries` / `state_grad_summaries` (empty when the
fragment sends nothing), `[N, fragments]` chains of flat slot indices padded with an identity
slot, the loss-cotangent source per reverse slot, and a terminal mask. `fold_slot_chains` is the
tensor form of the host fold: `L` batched merges per row regardless of real chain length.
`context_parallel_chunk(..., routing=, group=)` reads every index from those tensors, so a CUDA
Graph captured once replays for any document layout of the same fragment table after copying the
new layout's routing in place. The host-static helpers (`compose_entry_states`, ...) remain for
direct use; `context_parallel_kda` / `context_parallel_gdn` take `routing` instead of
`cu_seqlens` + `plan`.

Tests: plan tests updated to fragment slots; `routing()` tensors reproduce the host folds on three
layouts (CPU) and reject an undersized cap; the two-GPU distributed test now also captures one
graph (forward, NCCL gathers, backward) and replays it across three document layouts with fresh
inputs, matching the unsharded op each time. The example and docs use `routing`.

## Agent note

Third PR of the graph-replayable CP follow-up. Default env 231 passed; distributed 5 passed;
Mega env 71 passed, 4 xfail; zig-zag example with `--cuda-graph` passes.


drisspg:

PR #489, opened 2026-09-05
Split portable affine summaries into device-planned work items

drisspg:
## Agent note

The portable forward/reverse affine-summary kernels currently scan each whole range in one
serial chunk chain per head/column tile. Route them through the existing device work planner,
scan each item's chunk interval, and compose the partial maps in the appropriate direction.
The native SM100 path, public APIs, budget policy, and tile selectors are unchanged. Budget 1
still reads the original bounds and runs just the scan kernel.

This targets underfilled Hopper grids at low head counts, not a blanket Hopper speedup. On the
132-SM H100, H=1/2 select budgets 4/2, while H=4 selects budget 1. Tests now exercise portable
splits, partial tails, empty/padded rows, forced int64 addressing, and changed-layout graph replay.

**Draft pending SM100 runtime validation.** The B200 gpu-dev request stayed queued and was
cancelled; access to the existing GB200 host was denied. Review also needs to consider the
fragmented-range overhead below. No tile or budget-policy tuning is included in this PR.

## Performance

Full public-API subsystem timing, including device planning, partial allocation, scanning, and
composition: one 32K-token range at H=1 is **3.43× faster forward / 3.60× faster reverse**.
Unlike the earlier precomputed-subrange diagnostic, these measurements include the actual planner.

H100 80GB HBM3 (132 SMs), Python 3.13.13, PyTorch `2.15.0.dev20260905+cu132`, Triton 3.8.0,
CuTeDSL 4.7.1. Baseline `67ede83`; candidate uses this PR's kernels. B=1, T=32768, K=V=128,
BT=64, bf16 factors/FP32 gates, seed 2026. CUDA events over 20 calls after five warmups;
medians of three interleaved A/B rounds (AB, BA, AB), fixed/reused inputs, no cache flush or
clock/power changes. Post-run SM clocks were 1980 MHz. Compilation and setup are excluded;
eager numbers include Python dispatch gaps. These are summary microbenchmarks, not model throughput.

Latencies in µs, baseline → PR:

| Heads | Ranges | Forward | Reverse |
|---:|---|---:|---:|
| 1 | 1 × 32768 | 659 → 192 | 985 → 273 |
| 1 | 4096 / 20480 / 8192 | 383 → 200 | 612 → 371 |
| 1 | 8 × 4096 | 101 → 126 | 245 → 274 |
| 2 | 1 × 32768 | 788 → 445 | 969 → 514 |
| 2 | 4096 / 20480 / 8192 | 489 → 347 | 608 → 588 |
| 2 | 8 × 4096 | 227 → 260 | 467 → 497 |
| 4 | 1 × 32768 | 746 → 746 | 932 → 932 |
| 4 | 4096 / 20480 / 8192 | 502 → 504 | 893 → 892 |
| 4 | 8 × 4096 | 350 → 350 | 916 → 916 |

**Trade-off:** H=1/2 with eight short ranges incur roughly 25–33 µs extra planner/compose overhead;
the unchanged policy does not bypass that work based on the number/content of ranges. H=4 remains
on its original single-launch specialization. Small-percent differences are not claimed as wins.

Single-range CUDA Graph replay, excluding capture (fwd / rev, µs):
- H=1: 658 / 982 → 187 / 267
- H=2: 782 / 965 → 438 / 507
- H=4: 742 / 929 → 744 / 930

Profiler traces confirmed the planner → portable scan → compose kernels at H=1/2 and scan-only
at H=4. Budget-one outputs were bit-exact against the archived baseline launchers for bf16/fp16,
including empty ranges and partial tails; all nine long-sequence head/layout combinations passed
pointwise and relative-RMS parity in both directions.

## Test Plan

H100 nightly environment, independent editable install verified from outside the checkout:

```bash
gpu-run --timeout 30 auto -- timeout --signal=TERM --kill-after=5s 300s .venv/bin/pytest -q -n 6 test/test_delta_affine_summary.py test/test_delta_rule_stages.py test/test_context_parallel_plan.py -p no:cacheprovider
gpu-run --gpus 2 --timeout 30 auto -- timeout --signal=TERM --kill-after=5s 300s .venv/bin/pytest -q -n 2 test/test_context_parallel_distributed.py -p no:cacheprovider
```

- **335 passed, 76 skipped** in the affine-summary/stages/planner set.
- **14 passed** in distributed/public recipe coverage, including two-rank graph replay.
- Regression check against pre-change Hopper implementation: **8 failed**, proving the new tests
  detect bypassed planning; the same targeted set then **8 passed, 6 skipped** with this change.
- After review made the split test independent of physical SM count: **2 passed, 2 skipped**.
- Original-kernel parity script: 15 cases, each checking both directions; budget-one comparisons
  use `atol=rtol=0`. Split comparisons use `atol=rtol=2e-5` plus FP32-relative-RMS ≤64 eps.
- `ruff check`, changed-file `ruff format --check`, `git diff --check`, and all applicable
  changed-file `prek` hooks passed.
- SM100 execution remains unverified. Public staged summaries remain eager/CUDA-Graph APIs;
  no new `torch.compile`/export/autograd contract is introduced.

<details>
<summary>Benchmark reproduction scripts</summary>

The source trees at `/home/dev/hopper-baseline` and `/home/dev/attention-gym-hopper` are
archives of base and PR respectively. Both runs use the candidate environment, with imports
verified via the printed source path. Save the Python driver as `/home/dev/pr_benchmark.py`
and the runner as `/home/dev/run_pr_benchmark.sh`, then run:

```bash
gpu-run --timeout 30 auto -- bash /home/dev/run_pr_benchmark.sh
```

```python
"""Matched public-API affine-summary A/B, run with PYTHONPATH selecting one source tree."""
from __future__ import annotations

import json
import sys
from functools import partial
from pathlib import Path

import torch
import triton

import attn_gym
from attn_gym.linear._delta_rule.cute.affine_summary_fwd import build_state_summaries
from attn_gym.linear._delta_rule.cute.affine_summary_rev import build_state_grad_summaries

sys.path.insert(0, str(Path(attn_gym.__file__).resolve().parents[1] / "test"))
from test_delta_affine_summary import make_summary_inputs


def measure(fn) -> float:
    """Mean CUDA-event latency of 20 calls after five warmed calls, in microseconds."""
    for _ in range(5):
        fn()
    torch.cuda.synchronize()
    start, end = torch.cuda.Event(enable_timing=True), torch.cuda.Event(enable_timing=True)
    start.record()
    for _ in range(20):
        fn()
    end.record()
    end.synchronize()
    return start.elapsed_time(end) * 1000 / 20


def main() -> None:
    """Record full planner/scan/compose timing; graphs exclude capture and warmup."""
    label, round_id = sys.argv[1:]
    props = torch.cuda.get_device_properties(0)
    print(json.dumps({"label": label, "round": round_id, "torch": torch.__version__,
                      "triton": triton.__version__, "source": attn_gym.__file__,
                      "gpu": props.name, "uuid": str(props.uuid), "sms": props.multi_processor_count}), flush=True)
    layouts = {"one": [(0, 32768)], "uneven": [(0, 4096), (4096, 24576), (24576, 32768)],
               "eight": [(i * 4096, (i + 1) * 4096) for i in range(8)]}
    for heads in (1, 2, 4):
        qg, kg, w, u, dout, aqk, gate = make_summary_inputs(2026, tokens=32768, heads=heads)
        for name, rows in layouts.items():
            bounds = torch.tensor(rows, dtype=torch.int32, device="cuda")
            fns = {"fwd": partial(build_state_summaries, kg, w, u, gate, bounds),
                   "rev": partial(build_state_grad_summaries, qg, kg, w, dout, aqk, gate, 128**-0.5, bounds)}
            for direction, fn in fns.items():
                actual = fn()
                assert torch.isfinite(actual).all()
                record = {"label": label, "round": round_id, "heads": heads,
                          "layout": name, "direction": direction, "eager_us": measure(fn)}
                if name == "one":
                    graph = torch.cuda.CUDAGraph()
                    with torch.cuda.graph(graph):
                        captured = fn()
                    graph.replay()
                    torch.testing.assert_close(captured, actual, rtol=0, atol=0)
                    record["graph_us"] = measure(graph.replay)
                    if round_id == "0":
                        with torch.profiler.profile(activities=[torch.profiler.ProfilerActivity.CPU, torch.profiler.ProfilerActivity.CUDA]) as prof:
                            fn()
                            torch.cuda.synchronize()
                        record["kernels"] = sorted({event.name for event in prof.events() if event.device_type == torch.autograd.DeviceType.CUDA})
                print(json.dumps(record), flush=True)


if __name__ == "__main__":
    main()
```

```bash
#!/usr/bin/env bash
set -euo pipefail
cd /tmp
for round in 0 1 2; do
    variants=(baseline candidate)
    if [[ "$round" == 1 ]]; then variants=(candidate baseline); fi
    for variant in "${variants[@]}"; do
        source_dir=/home/dev/hopper-baseline
        if [[ "$variant" == candidate ]]; then source_dir=/home/dev/attention-gym-hopper; fi
        PYTHONPATH="$source_dir" timeout --signal=TERM --kill-after=5s 60s /home/dev/attention-gym-hopper/.venv/bin/python -P /home/dev/pr_benchmark.py "$variant" "$round"
        nvidia-smi --query-gpu=uuid,clocks.sm,power.draw,temperature.gpu --format=csv,noheader
    done
done
```

</details>

### DIFF
--- a/attn_gym/linear/_delta_rule/triton/affine_summary_fwd.py
+++ b/attn_gym/linear/_delta_rule/triton/affine_summary_fwd.py
@@ -6,8 +6,8 @@
 
 """Portable Triton forward affine summaries for delta-rule context parallelism.
 
-Each program keeps one FP32 ``[128, BN]`` augmented-state tile resident while
-scanning complete BT64 chunks. Raw compact-layout pointers avoid host tensor
+Each program scans one work item (row layout in ``work_items.py``), keeping one
+FP32 ``[128, BN]`` augmented-state tile resident while scanning BT64 chunks. Raw compact-layout pointers avoid host tensor
 descriptors, allocations, and synchronization in the launch path, so a warmed
 specialization can be captured directly by a CUDA Graph.
 """
@@ -43,39 +43,54 @@ def affine_summary_fwd_kernel(
     w,
     u,
     cumulative_gate,
-    bounds,
+    work,
     out,
     H: tl.constexpr,
     K: tl.constexpr,
     V: tl.constexpr,
     BT: tl.constexpr,
     BN: tl.constexpr,
+    WHOLE_RANGES: tl.constexpr,
     USE_INT64_OFFSETS: tl.constexpr,
 ):
-    """Scan one range, head, and augmented-state column tile through its BT64 chunks.
+    """Scan one work item, head, and augmented-state column tile through its BT64 chunks.
 
-    ``bounds[range] = (start, stop)`` are token offsets; the partial last chunk is masked so
-    tokens past ``stop`` contribute nothing and its gate row is the range's final token.
+    ``work_items.py`` defines the work-row layout. The partial last chunk is masked so
+    tokens past the range's end contribute nothing; unused rows store the identity.
     """
     head = tl.program_id(0)
     column_tile = tl.program_id(1)
-    range_id = tl.program_id(2)
+    work_index = tl.program_id(2)
     if USE_INT64_OFFSETS:
         head = head.to(tl.int64)
         column_tile = column_tile.to(tl.int64)
-        range_id = range_id.to(tl.int64)
-    start = tl.load(bounds + 2 * range_id)
-    stop = tl.load(bounds + 2 * range_id + 1)
-    if USE_INT64_OFFSETS:
-        start = start.to(tl.int64)
-        stop = stop.to(tl.int64)
+        work_index = work_index.to(tl.int64)
+    if WHOLE_RANGES:
+        start = tl.load(work + 2 * work_index)
+        stop = tl.load(work + 2 * work_index + 1)
+        if USE_INT64_OFFSETS:
+            start = start.to(tl.int64)
+            stop = stop.to(tl.int64)
+        chunk_begin = 0
+        chunk_end = (stop - start + BT - 1) // BT
+    else:
+        start = tl.load(work + 4 * work_index)
+        chunk_begin = tl.load(work + 4 * work_index + 1)
+        chunk_end = tl.load(work + 4 * work_index + 2)
+        length = tl.load(work + 4 * work_index + 3)
+        if USE_INT64_OFFSETS:
+            start = start.to(tl.int64)
+            chunk_begin = chunk_begin.to(tl.int64)
+            chunk_end = chunk_end.to(tl.int64)
+            length = length.to(tl.int64)
+        stop = start + length
 
     key = tl.arange(0, K)
     row = tl.arange(0, BT)
     column = column_tile * BN + tl.arange(0, BN)
     state = tl.where(column[None, :] == V + key[:, None], 1.0, 0.0).to(tl.float32)
 
-    for chunk in tl.range(0, (stop - start + BT - 1) // BT):
+    for chunk in tl.range(chunk_begin, chunk_end):
         token_start = start + chunk * BT
         token = token_start + row
         valid = token[:, None] < stop
@@ -112,7 +127,7 @@ def affine_summary_fwd_kernel(
         state = tl.dot(tl.trans(kg_tile), tmp_lo, acc=state)
 
     out_offset = ptr_offset(
-        (range_id, head, column[None, :], key[:, None]),
+        (work_index, head, column[None, :], key[:, None]),
         (H * (V + K) * K, (V + K) * K, K, 1),
     )
     tl.store(out + out_offset, state)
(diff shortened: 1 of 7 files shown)


attn_gym/linear/_delta_rule/triton/affine_summary_fwd.py
```diff
@@ -43,39 +43,54 @@ def affine_summary_fwd_kernel(
...
+    else:
+        start = tl.load(work + 4 * work_index)
+        chunk_begin = tl.load(work + 4 * work_index + 1)
+        chunk_end = tl.load(work + 4 * work_index + 2)
+        length = tl.load(work + 4 * work_index + 3)
+        if USE_INT64_OFFSETS:
+            start = start.to(tl.int64)
+            chunk_begin = chunk_begin.to(tl.int64)
+            chunk_end = chunk_end.to(tl.int64)
+            length = length.to(tl.int64)
```
drisspg: can we turn this hlerp somwehrw next to theowrk chunking kind of similar ot what ihave for the raggedmetadat stuff

PR #548, opened 2026-09-14
add bi and bwd for recurrent kernel

liangel-02:
instead of using disable autotune as bi flag, i define a separate bi bool because we also need to disable the heuristic that selects bv and the number of warps 

also i added bwd support for the recurrent kernel so that for batch invariance ppl can use recurrent kernel for training + inference


attn_gym/linear/_delta_rule/recurrent.py
```diff
@@ -202,6 +202,7 @@ def launch_recurrent_delta_rule_fwd(
     state_indices: torch.Tensor | None = None,
     has_initial_state: torch.Tensor | None = None,
     autotune: bool = False,
+    batch_invariant: bool = False,
```
drisspg, line 205: lets put into kenrel options

PR #549, opened 2026-09-15
add support for gdn head dim = 64

liangel-02:
for non cudnn path

use triton autotuning for local dv, reverse delta h , q/k/w grads, wy bwd and w/u/kg construction. delta h is most performant at bk = 64, bv = 32, warps = 4.  
Screenshot 2026-09-15 at 12 59 44 PM


attn_gym/linear/gdn/bwd/triton/chunk_gdn_bwd_delta_h.py
```diff
@@ -168,17 +199,19 @@ def chunk_gdn_bwd_delta_h_kernel(
     row = tl.arange(0, BT)
     value = value_tile * BV + tl.arange(0, BV)
     key_1 = tl.arange(0, BK)
-    key_2 = BK + key_1
     key_head = head // GROUPS
     d_state_1 = tl.zeros((BK, BV), dtype=tl.float32)
-    d_state_2 = tl.zeros((BK, BV), dtype=tl.float32)
+    if K > BK:
+        key_2 = BK + key_1
+        d_state_2 = tl.zeros((BK, BV), dtype=tl.float32)
```
drisspg, line 206: Could we express these key-tile paths as compile-time-unrolled loops instead of maintaining separate `d_state_1` / `d_state_2` branches throughout the recurrence? Keep `BV` as the independent parallel tiling axis, represent the resident state tiles with a Triton-supported compile-time structure, then use `tl.static_range(tl.cdiv(K, BK))` to (1) accumulate all key-tile contributions to `value_gradient` and (2) update each state tile once that gradient is complete. The state tiles should remain live across chunk iterations, rather than introducing global-memory scratch. With key/value tail masks, this would give us a path to other K/V sizes without adding more numbered branches. We'd need to verify Triton's lowering and performance, but I'd prefer that structure over hard-coding the one-/two-tile cases.

drisspg, line 206: Concretely, could we replace the numbered state variables with a compile-time tuple and statically unroll the key-tile work? Below are coordinated replacement excerpts, not a drop-in replacement for just these three lines. They reuse the existing per-chunk offsets/decays.

**1. Initialize the resident tiles before the reverse chunk loop:**

```python
NK: tl.constexpr = triton.cdiv(K, BK)
key_lane = tl.arange(0, BK)
value_mask = value < V
state_base = ptr_offset((sequence, head), (H * V * K, V * K))
d_states = ()
for ki in tl.static_range(NK):
    key = ki * BK + key_lane
    state = tl.zeros((BK, BV), dtype=tl.float32)
    if USE_FINAL_STATE_GRADIENT:
        offset = state_base + value[None, :] * K + key[:, None]
        state += tl.load(
            d_final_state + offset,
            mask=(key[:, None] < K) & value_mask[None, :],
            other=0.0,
        ).to(tl.float32)
    d_states = d_states + (state,)
```

**2. Inside each chunk, replace the manually appended key contributions:**

```python
value_gradient = tl.zeros((BT, BV), dtype=tl.float32)
for ki in tl.static_range(NK):
    key = ki * BK + key_lane
    k_tile = tl.load(
        k + k_base + key[None, :],
        mask=token_mask[:, None] & (key[None, :] < K),
        other=0.0,
    )
    value_gradient += tl.dot(k_tile, d_states[ki].to(k_tile.dtype))

value_gradient *= restored_decay[:, None]
value_gradient += tl.load(
    dv_local + value_offset,
    mask=token_mask[:, None] & value_mask[None, :],
    other=0.0,
)
tl.store(
    dv + value_offset,
    value_gradient.to(dv.dtype.element_ty),
    mask=token_mask[:, None] & value_mask[None, :],
)
```

**3. Once `value_gradient` is complete, rebuild the state tuple for the next chunk:**

```python
output_tile = tl.load(
    d_output + value_offset,
    mask=token_mask[:, None] & value_mask[None, :],
    other=0.0,
)
next_states = ()
for ki in tl.static_range(NK):
    key = ki * BK + key_lane
    mask = token_mask[:, None] & (key[None, :] < K)
    q_tile = tl.load(q + q_base + key[None, :], mask=mask, other=0.0)
    w_tile = tl.load(w + w_base + key[None, :], mask=mask, other=0.0)
    q_tile = (q_tile * query_decay[:, None]).to(q_tile.dtype)

    state = d_states[ki] * state_decay
    state += tl.dot(tl.trans(q_tile), output_tile) * scale
    state -= tl.dot(tl.trans(w_tile), value_gradient.to(w_tile.dtype))
    next_states = next_states + (state,)
d_states = next_states
```

Keep the existing definitions of `k_base`, `q_base`, `w_base`, `value_offset`, `state_decay`, and `query_decay` before their respective uses. The `dh` boundary stores (before updating state) and optional final `d_initial_state` stores should use the same static key loop and key/value masks. `BV` remains the parallel grid axis; there is no explicit global-memory state scratch here.

I'd first validate this as a structural refactor on the existing 64/128 cases. The tuple construction/indexing and loop-carried tuple must compile on the supported Triton version, and we should check spills/performance rather than assume equivalent lowering. I haven't compiled this proposal. Broadening K/V support is a follow-up: remove the `K <= 2 * BK` pruning constraint, update the wrapper restrictions, and add key/value tail masks to the separate `dv_local` kernel too—not just this recurrence.

attn_gym/linear/gdn/fwd/triton/chunk_gdn_fwd_recurrence.py
```diff
@@ -76,15 +77,21 @@ def chunk_gdn_fwd_recurrence_dense(
...
     if initial_state.shape != (batch, heads, value_dim, key_dim):
         raise ValueError("initial_state must have shape [B,H,V,K]")
 
     chunks = tokens // 64
-    block_value = 64
+    # Split D64 state columns while the grid needs occupancy; once chunk-head work is ample,
+    # BV64 avoids duplicating the serial K/W traffic across two programs.
+    use_small_value_tile = value_dim == 64 and (tokens // 64) * heads < 8192
+    block_value = 32 if use_small_value_tile else 64
+    num_stages = 2 if use_small_value_tile else 3
```
drisspg, line 94: Can we justify this selector using the actual launch geometry? The grid is `(batch * heads, V // BV)`; chunks execute serially inside each program, so increasing `T` doesn't supply more parallel programs. If `8192` is a measured crossover between value-tile parallelism and duplicated K/W traffic, could we record the GPU, workload sweep, and timings and describe it that way? Otherwise I'd base an occupancy heuristic on the program count and device rather than `chunks * heads`.

test/linear/test_gdn_chunk_fused.py
```diff
@@ -204,18 +236,32 @@ def test_fused_chunk_matches_low_precision_reference(
...
-        if high_precision.abs().max().item() < 1e-12:
+        if lengths is not None and head_dim == 64 and name == "dstate":
+            # The entry-state cotangent crosses the QKV-dtype chunk checkpoint.
+            # Account for that additional rounding boundary over the eager oracle.
+            high = high_precision.double()
+            actual_error = (result.double() - high).abs().max().item()
+            reference_error = (reference.double() - high).abs().max().item()
+            rounding = torch.finfo(dtype).eps * high.abs().max().item()
+            assert actual_error <= 4 * (reference_error + rounding)
+        elif high_precision.abs().max().item() < 1e-12:
```
drisspg, line 247: Could we explain the three D64-only numerical adjustments together: scaling packed initial states by `0.01`, switching the state loss from mean to sum, and increasing this budget from 2× to 4×? I'd retain a common input/loss baseline across dimensions and add a separate scaling/underflow case if needed. The initial-state gradient is stored directly from FP32 accumulators, so this comment should identify the actual low-precision operation and why its effect warrants a different budget at D64. Please include the observed kernel/reference errors motivating the allowance rather than just the dimension-specific exception.

test/linear/test_gdn_chunk_fused.py
```diff
@@ -143,21 +148,37 @@ def force_portable_backward(monkeypatch) -> None:
...
+        (2, 65, 1, 4, 128, torch.bfloat16, "mild", None, False),
+        (1, 64, 1, 1, 128, torch.bfloat16, "unbounded", None, True),
+        (1, 64, 1, 64, 128, torch.bfloat16, "mild", None, True),
+        (1, 192, 1, 4, 128, torch.bfloat16, "spikes", [65, 0, 127], True),
+        (1, 64, 1, 2, 64, torch.bfloat16, "mild", None, True),
+        (1, 64, 1, 2, 64, torch.float16, "mild", None, True),
+        (1, 16, 1, 2, 64, torch.bfloat16, "mild", [7, 9], True),
+        (1, 16, 1, 2, 64, torch.float16, "mild", [7, 9], True),
     ],
 )
```
drisspg, line 174: Can we add a multi-chunk D64 case to this independent-oracle test? These D64 rows only cover one complete chunk or short partial chunks, all with mild gates. Mirroring the existing D128 row would exercise the backward recurrence across chunks, a partial tail, and empty-sequence state identity:

```python
(1, 192, 1, 4, 64, torch.bfloat16, "spikes", [65, 0, 127], True),
```

The compiled-versus-eager fused tests are useful, but don't independently validate the recurrence's numerical result.

attn_gym/linear/gdn/bwd/triton/chunk_gdn_bwd_wy.py
```diff
@@ -569,6 +594,7 @@ def chunk_gdn_bwd_wy(
             tokens,
             0 if metadata is None else metadata.cu_seqlens.shape[0] - 1,
             H=value_heads,
+            KEY_TILES=key_dim // block_key,
```
drisspg, line 597: For the broader K support discussed on the state recurrence, could we derive `key_tiles = triton.cdiv(key_dim, block_key)` once and reuse it for the `d_gate_direct` allocation, `direct_grid`, and this argument? All three currently use floor division. This also needs `keys < K` masks on the direct kernel's Q/K loads and dQ/dK/dW stores; ceiling division alone would launch an unsafe partial tile. The current 64/128 contract is safe—this is a coordinated part of generalizing it, not a current tail bug.

attn_gym/linear/gdn/fwd/triton/chunk_gdn_fwd_recurrence.py
```diff
@@ -76,15 +77,24 @@ def chunk_gdn_fwd_recurrence_dense(
...
     if initial_state.shape != (batch, heads, value_dim, key_dim):
         raise ValueError("initial_state must have shape [B,H,V,K]")
 
     chunks = tokens // 64
-    block_value = 64
+    # Chunks execute serially inside each program, so only batch/head/value tiles contribute
+    # launch parallelism. Split D64 values when the unsplit grid cannot fill one GPU wave;
+    # otherwise BV64 avoids duplicating the K/W traffic across two programs.
+    program_count = batch * heads
+    num_sms = torch.cuda.get_device_properties(k.device).multi_processor_count
```
drisspg: Can we reuse the cached `get_device_properties` utility from `attn_gym._backends.cute.utils` here instead of calling `torch.cuda.get_device_properties` directly? Other Triton code already uses it (e.g. `chunk_kda_fwd_intra_sub_chunk_forloop.py`), so this can be `get_device_properties(k.device).multi_processor_count`.

attn_gym/linear/gdn/bwd/triton/chunk_gdn_bwd_delta_h.py
```diff
@@ -167,18 +208,23 @@ def chunk_gdn_bwd_delta_h_kernel(
...
-        final_offset_1 = state_base + ptr_offset((value[:, None], key_1[None, :]), (K, 1))
-        final_offset_2 = state_base + ptr_offset((value[:, None], key_2[None, :]), (K, 1))
-        d_state_1 += tl.trans(tl.load(d_final_state + final_offset_1)).to(tl.float32)
-        d_state_2 += tl.trans(tl.load(d_final_state + final_offset_2)).to(tl.float32)
+    d_states = ()
+    for ki in tl.static_range(NK):
+        key = ki * BK + key_lane
+        state = tl.zeros((BK, BV), dtype=tl.float32)
+        if USE_FINAL_STATE_GRADIENT:
+            offset = state_base + value[None, :] * K + key[:, None]
```
drisspg:
```suggestion
            offset = state_base + ptr_offset((value[None, :], key[:, None]), (K, 1))
```

attn_gym/linear/gdn/bwd/triton/chunk_gdn_bwd_delta_h.py
```diff
@@ -208,75 +258,68 @@ def chunk_gdn_bwd_delta_h_kernel(
...
+        d_states = next_states
 
     if STORE_INITIAL_STATE_GRADIENT:
-        initial_offset_1 = state_base + ptr_offset((value[:, None], key_1[None, :]), (K, 1))
-        initial_offset_2 = state_base + ptr_offset((value[:, None], key_2[None, :]), (K, 1))
-        tl.store(d_initial_state + initial_offset_1, tl.trans(d_state_1))
-        tl.store(d_initial_state + initial_offset_2, tl.trans(d_state_2))
+        for ki in tl.static_range(NK):
+            key = ki * BK + key_lane
+            offset = state_base + value[None, :] * K + key[:, None]
```
drisspg:
```suggestion
            offset = state_base + ptr_offset((value[None, :], key[:, None]), (K, 1))
```

attn_gym/linear/gdn/bwd/triton/chunk_gdn_bwd_delta_h.py
```diff
@@ -188,10 +234,14 @@ def chunk_gdn_bwd_delta_h_kernel(
         token_mask = row < valid_tokens
 
         dh_base = ptr_offset((global_chunk, head), (H * K * V, K * V))
-        dh_offset_1 = dh_base + ptr_offset((key_1[:, None], value[None, :]), (V, 1))
-        dh_offset_2 = dh_base + ptr_offset((key_2[:, None], value[None, :]), (V, 1))
-        tl.store(dh + dh_offset_1, d_state_1.to(dh.dtype.element_ty))
-        tl.store(dh + dh_offset_2, d_state_2.to(dh.dtype.element_ty))
+        for ki in tl.static_range(NK):
+            key = ki * BK + key_lane
+            dh_offset = dh_base + key[:, None] * V + value[None, :]
```
drisspg:
```suggestion
            dh_offset = dh_base + ptr_offset((key[:, None], value[None, :]), (V, 1))
```

drisspg (approved):
Some small commetns, the main thing, lets do in a follow up is that I dont get why we are restircted to such a smalls of swquare state and only 64x128, if we can cheapily and simply expand to supproting more state shapes we should

PR #550, opened 2026-09-15
add bf16 support to gdn and kda

liangel-02:
adding selectable persistent state-cache precision for inference bf16 and fp32, this is for gdn and kda for the atteniton gym backend not for cudnn 
parameterizing existing tests to cover fp32 and bf16


test/linear/test_gdn_fused.py
```diff
@@ -165,7 +171,7 @@ def test_fused_recurrent_packed_paged_state():
             if begin == end or slot <= 0:
                 continue
             span = slice(begin, end)
-            initial_state = original_cache[slot].unsqueeze(0) if use_state else None
+            initial_state = original_cache[slot].unsqueeze(0).float() if use_state else None
```
drisspg: should we update the ref to also support bf16?

drisspg (approved):

PR #556, opened 2026-09-16
paged chunked kda decode for batch invariance

liangel-02:
for batch invariance you can either match decode to prefill (recurrent/recurrent) or match prefill to decode (chunked/chunked) and we found through running benchmarking w/ titan's grpo loop that its faster to match chunked/chunked but only if we remove the cache wide copies due to functionalized indexed cache updates. so this pr extends the existing paged_chunk_kda interface with replay-backed decode to support inplace writes direclty into vllm cache slots for decode

image


attn_gym/linear/kda/api.py
```diff
@@ -195,6 +257,15 @@ def paged_chunk_kda(
     has_initial_state: torch.Tensor | None = None,
     autotune: bool = True,
     kernel_options: KernelOptions | None = None,
+    replay_state: tuple[
```
drisspg: nit: lets define a type at the top

attn_gym/linear/kda/api.py
```diff
@@ -45,6 +45,68 @@
...
+    ):
+        if tensor.shape != expected_shape:
+            raise ValueError(
+                f"replay_state.{name} must have shape {expected_shape}, got {tuple(tensor.shape)}"
+            )
+        if tensor.dtype != expected_dtype or tensor.device != q.device:
+            raise TypeError(f"replay_state.{name} must use {expected_dtype} on the input device")
+        expected_stride = 1
+        dense_inner = True
+        for size, stride in reversed(tuple(zip(tensor.shape[2:], tensor.stride()[2:]))):
```
drisspg: Let's change the error messages so they don't imply `replay_state` has named attributes—it's a tuple. Something like “replay_state q tensor” would be clearer than “replay_state.q”.

attn_gym/linear/kda/impl/fused.py
```diff
@@ -260,22 +270,168 @@ def paged_chunk_forward(
...
+            state_indices,
+            *replay_state,
+            has_initial_state=has_initial_state,
+            scale=q.shape[-1] ** -0.5,
+            autotune=autotune,
+            schedule=schedule,
+        )
+
+    output_dtype = q.dtype
+    q, k, v = (tensor.to(torch.bfloat16).contiguous() for tensor in (q, k, v))
```
drisspg: Can we drop `.contiguous()` here and on the following gate/beta casts? This path only slices these inputs and copies them into the replay buffers with `copy_`, which supports noncontiguous sources. The kernel consumes the replay buffers, so these extra copies don't seem necessary.

attn_gym/linear/kda/impl/fused.py
```diff
@@ -260,22 +270,168 @@ def paged_chunk_forward(
...
+            (0, int(cu_seqlens[index].item()), int(cu_seqlens[index + 1].item()))
+            for index in range(state_indices.numel())
+        ]
+
+    for sequence_index, start, end in sequence_ranges:
+        slot = int(state_indices[sequence_index].item())
+        if slot <= 0:
+            continue
+        fresh = has_initial_state is not None and not bool(
+            has_initial_state[sequence_index].item()
```
drisspg: Could we separate the logical sequence index from the physical batch index here? Packed ranges all start with `0`, which correctly selects the token batch row but also makes every sequence read `state_indices[0]` and `has_initial_state[0]`.

For example, with `cu_seqlens=[0, 64, 128]`, `state_indices=[3, 1]`, and `has_initial_state=[True, False]`, A should reuse slot 3 and B should start fresh in slot 1. Instead, both use slot 3 and the first initialization flag; slot 1 is untouched. This silently produces incorrect state/output rather than necessarily raising an exception. An especially simple failing case is routes `[0, 1]`: only A is padding, but both sequences are skipped and the entire output is zero.

Please add a packed replay regression against independent per-sequence calls, with distinct slots and initialization flags, plus the leading-padding case.

drisspg: Could we also make this replay prefill/packed path CUDA-graphable? The `.item()` calls here and below read CUDA offsets, slot indices, initialization flags, and replay counts into Python. Those device-to-host reads require synchronization and fail during CUDA graph capture; warming up first or replacing `.item()` with `int(tensor)` won't fix it. The extracted values also drive Python branches and loop lengths, so routing/reset/count handling needs to stay on-device with a capture-stable launch structure. Please add a CUDA graph capture/replay regression for this path, including changed routing/count values between replays. This is separate from the logical-vs-physical index bug; dense single-token decode bypasses this loop.

attn_gym/linear/kda/impl/fused.py
```diff
@@ -306,4 +462,106 @@ def paged_chunk_forward(
...
+            replay_beta,
+        )
+    ):
+        raise RuntimeError(
+            "paged_chunk_kda replay mode is inference-only; call under torch.no_grad() or "
+            "torch.inference_mode()"
+        )
+
+    output_dtype = q.dtype
+    q, k, v = (tensor.to(torch.bfloat16).contiguous() for tensor in (q, k, v))
```
drisspg: same deal why all these contig calls

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,290 @@
...
+    replay_cache,
+    current,
+    state_indices,
+    active_counts,
+    active_cache,
+    cache_stride_s: tl.constexpr,
+    cache_stride_t: tl.constexpr,
+    current_stride_b: tl.constexpr,
+    active_stride_b: tl.constexpr,
+    active_stride_t: tl.constexpr,
```
drisspg:
```suggestion
    cache_strides: tl.constexpr,
    current_strides: tl.constexpr,
    active_strides: tl.constexpr,
```

drisspg (commented):
i need to think about cuda-graphing this

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,290 @@
...
+    is_current = active & (token == count)
+    current_value = tl.load(
+        current + active_index * current_stride_b + feature,
+        mask=is_current,
+        other=0.0,
+    )
+    value = tl.where(is_current, current_value, value)
+    tl.store(replay_cache + cache_offset, value, mask=is_current)
+    active_offset = active_index * active_stride_b + token * active_stride_t + feature
+    tl.store(active_cache + active_offset, value, mask=in_bounds)
```
drisspg:
```suggestion
    cache_offset = slot * cache_strides[0] + token * cache_strides[1] + feature
    current_offset = active_index * current_strides[0] + feature
    active_offset = active_index * active_strides[0] + token * active_strides[1] + feature
    cache_ptr = replay_cache + cache_offset
    current_ptr = current + current_offset
    active_ptr = active_cache + active_offset

    value = tl.load(cache_ptr, mask=active, other=0.0)
    is_current = active & (token == count)
    current_value = tl.load(current_ptr, mask=is_current, other=0.0)
    value = tl.where(is_current, current_value, value)
    tl.store(cache_ptr, value, mask=is_current)
    tl.store(active_ptr, value, mask=in_bounds)
```

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,290 @@
...
+            cache,
+            current,
+            state_indices,
+            active_counts,
+            active_cache,
+            cache_stride_s=cache.stride(0),
+            cache_stride_t=cache.stride(1),
+            current_stride_b=current.stride(0),
+            active_stride_b=active_cache.stride(0),
+            active_stride_t=active_cache.stride(1),
```
drisspg:
```suggestion
            cache_strides=cache.stride(),
            current_strides=current.stride(),
            active_strides=active_cache.stride(),
```

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,290 @@
...
+    feature = offset % feature_size
+    remaining = offset // feature_size
+    token = remaining % num_tokens
+    active_index = remaining // num_tokens
+
+    slot = tl.load(state_indices + active_index, mask=in_bounds, other=0).to(tl.int64)
+    active = in_bounds & (slot > 0)
+    count = tl.load(active_counts + active_index, mask=in_bounds, other=0)
+    cache_offset = slot * cache_stride_s + token * cache_stride_t + feature
+    value = tl.load(replay_cache + cache_offset, mask=active, other=0.0)
```
drisspg:
```suggestion
    value = tl.load(
        replay_cache + cache_offset, mask=active & (token < count), other=0.0
    )
```

drisspg: Can we mask this load with `active & (token < count)` and keep `other=0.0`? We already visit every element of every active replay window, and the code below inserts the current token at `token == count`, so this would zero the unused tail in `active_cache` without another pass or clearing the persistent cache. Currently the unused tail is read verbatim; NaNs/Infs there could contaminate valid outputs through downstream matmuls despite causal masking (`0 * NaN`). A test that poisons the unused replay tail with NaNs would cover this—the existing dirty-cache test uses finite random values.

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,290 @@
...
+    cache_stride_s: tl.constexpr,
+    cache_stride_t: tl.constexpr,
+    current_stride_b: tl.constexpr,
+    active_stride_b: tl.constexpr,
+    active_stride_t: tl.constexpr,
+    num_tokens: tl.constexpr,
+    feature_size: tl.constexpr,
+    num_elements: tl.constexpr,
+    BLOCK: tl.constexpr,
+):
```
drisspg, line 388: Can we group the strides into per-tensor tuples and use the existing `ptr_offset` utility from `attn_gym._backends.triton.utils` for the cache/current/active offsets instead of spelling out the arithmetic? For example:

```python
cache_offset = ptr_offset(
    (slot, token, feature), (cache_strides[0], cache_strides[1], 1)
)
```

The trailing `1` preserves the existing contiguous flattened-feature assumption; the full tensor stride tuple doesn't correspond directly to these flattened indices. This supersedes the manual offset arithmetic in my earlier suggestion.

attn_gym/linear/kda/impl/fused.py
```diff
@@ -260,22 +270,168 @@ def paged_chunk_forward(
     has_initial_state: torch.Tensor | None = None,
     autotune: bool = True,
     schedule: ScheduleRequest = ScheduleRequest.AUTO,
+    replay_state: _ReplayState | None = None,
 ) -> torch.Tensor:
-    """Normalize inputs and invoke the registered paged chunk operator."""
+    """Run paged chunk prefill or replay-backed decode."""
     _validate_fused_constraints(q, v)
+    replay_tensors = () if replay_state is None else replay_state[:-1]
```
drisspg: let sjsut remove -1 it was a lil confusign at afirst

attn_gym/linear/kda/impl/fused.py
```diff
@@ -260,22 +270,168 @@ def paged_chunk_forward(
...
+    output = torch.zeros_like(v, dtype=output_dtype)
+    if cu_seqlens is None:
+        sequence_ranges = [(index, 0, q.shape[1]) for index in range(q.shape[0])]
+    else:
+        sequence_ranges = [
+            (0, int(cu_seqlens[index].item()), int(cu_seqlens[index + 1].item()))
+            for index in range(state_indices.numel())
+        ]
+
+    for sequence_index, start, end in sequence_ranges:
```
drisspg:
```suggestion
    for sequence_index, (batch_index, start, end) in enumerate(sequence_ranges):
```

attn_gym/linear/kda/impl/fused.py
```diff
@@ -260,22 +270,168 @@ def paged_chunk_forward(
...
+        if fresh:
+            state_cache[slot].zero_()
+            replay_count[slot].zero_()
+        count = int(replay_count[slot].item())
+        cursor = start
+        while cursor < end:
+            num_tokens = min(_CHUNK_SIZE - count, end - cursor)
+            next_count = count + num_tokens
+            source = (
+                slice(sequence_index, sequence_index + 1),
```
drisspg:
```suggestion
                slice(batch_index, batch_index + 1),
```

attn_gym/linear/kda/impl/fused.py
```diff
@@ -260,22 +270,168 @@ def paged_chunk_forward(
...
+            state_cache,
+            state_indices,
+            *replay_state,
+            has_initial_state=has_initial_state,
+            scale=q.shape[-1] ** -0.5,
+            autotune=autotune,
+            schedule=schedule,
+        )
+
+    output_dtype = q.dtype
```
drisspg:
```suggestion
    if torch.compiler.is_compiling():
        raise RuntimeError(
            "Replay-backed multi-token or packed prefill is eager-only; "
            "torch.compile is supported only for dense single-token replay decode."
        )
    if torch.cuda.is_current_stream_capturing():
        raise RuntimeError(
            "Replay-backed multi-token or packed prefill is eager-only; "
            "CUDA capture is supported only for dense single-token replay decode."
        )

    output_dtype = q.dtype
```

attn_gym/linear/kda/api.py
```diff
@@ -222,6 +293,11 @@ def paged_chunk_kda(
             the optional CuTeDSL 4.7 cuDNN backend. ``split_backward`` and
             ``split_forward`` are not supported by this paged operation; ``schedule``
             applies to the fused backend as in ``chunk_kda``.
+        replay_state: Optional ``(q, k, v, gate, beta, count)`` persistent cache tuple.
+            Q, K, and V use BF16, gate and beta use FP32, and count uses int32. The five
+            token caches have shape ``[num_slots, 64, ...]``. When supplied, calls with one
+            token per dense batch row use replay-backed decode; other calls use streaming
+            prefill. Both paths preserve chunk boundaries across calls.
```
drisspg:
```suggestion
            prefill. Both paths preserve chunk boundaries across calls. Replay-backed
            streaming prefill is eager-only and synchronizes device metadata to the host;
            torch.compile and CUDA capture are supported only for dense single-token
            replay decode.
```

attn_gym/linear/kda/api.py
```diff
@@ -45,6 +45,68 @@
...
+            )
+        if tensor.dtype != expected_dtype or tensor.device != q.device:
+            raise TypeError(f"replay_state.{name} must use {expected_dtype} on the input device")
+        expected_stride = 1
+        dense_inner = True
+        for size, stride in reversed(tuple(zip(tensor.shape[2:], tensor.stride()[2:]))):
+            dense_inner &= size == 1 or stride == expected_stride
+            expected_stride *= size
+        if not dense_inner or tensor.stride(1) != expected_stride:
+            raise ValueError(f"replay_state.{name} must have dense per-token rows")
```
drisspg:
```suggestion
        if not dense_inner or tensor.stride(1) != expected_stride:
            raise ValueError(f"replay_state.{name} must have dense per-token rows")
        if num_slots > 1 and tensor.stride(0) < _CHUNK_SIZE * expected_stride:
            raise ValueError(f"replay_state.{name} slots must not overlap")
```

attn_gym/linear/kda/api.py
```diff
@@ -45,6 +45,68 @@
...
+    count = replay_state[-1]
+    if count.shape not in ((num_slots,), (num_slots, 1)):
+        raise ValueError(
+            f"replay_state.count must have shape ({num_slots},) or ({num_slots}, 1), "
+            f"got {tuple(count.shape)}"
+        )
+    if count.dtype != torch.int32 or count.device != q.device:
+        raise TypeError("replay_state.count must be int32 on the input device")
```
drisspg:
```suggestion
    if count.dtype != torch.int32 or count.device != q.device:
        raise TypeError("replay_state.count must be int32 on the input device")
    if num_slots > 1 and count.stride(0) == 0:
        raise ValueError("replay_state.count slots must not overlap")
```

attn_gym/linear/kda/impl/fused.py
drisspg: Apply both indexing suggestions together: enumerate supplies the logical sequence index for state_indices/has_initial_state, while batch_index stays zero for packed token tensors. Please add a packed replay regression with offsets [0, 32, 64], slots [2, 3], and has_initial_state=[False, True]. Give slot 3 a valid pending history that crosses a 64-token boundary after appending; compare outputs and all mutated cache tensors against independent dense calls. Check untouched slots too.

attn_gym/linear/kda/impl/fused.py
drisspg: This suggestion takes the bounded eager-only option, not an on-device prefill implementation. Reject before conversions, .item(), or cache mutation, and document the restriction publicly. Add rejection coverage for both dense multi-token and packed replay prefill under fullgraph compilation and CUDA capture, retaining the existing successful dense single-token compile/capture test. Dynamo may wrap the explicit exception as an Unsupported error; verify rejection rather than assuming the eager exception type is preserved.

attn_gym/linear/kda/api.py
drisspg: The suggested slot-stride guard preserves non-overlapping outer-strided pools rather than requiring contiguous caches. Please parameterize rejection tests across all five token caches using a one-slot tensor expanded across slots, plus expanded counts in both accepted shapes. Also test a positive but too-small token slot stride, and retain a positive case with gapped token slots and strided counts (the existing [:, 1:2] count layout must remain supported).

attn_gym/linear/kda/api.py
```diff
@@ -248,6 +259,64 @@ def paged_chunk_kda(
...
+            raise TypeError("replay_state count tensor must be int32 on the input device")
+        if num_slots > 1 and replay_state.count.stride(0) == 0:
+            raise ValueError("replay_state count tensor slots must not overlap")
+        valid_counts = ((replay_state.count >= 0) & (replay_state.count < _CHUNK_SIZE)).all()
+        if torch.compiler.is_compiling() or (
+            replay_state.count.is_cuda and torch.cuda.is_current_stream_capturing()
+        ):
+            torch._assert_async(valid_counts, "replay_state counts must be in [0, 63]")
+        elif not valid_counts.item():
+            raise ValueError("replay_state counts must be in [0, 63]")
```
drisspg: why do we have this werid like split; can we jsut not gaurd agaisnt thsi im okay with assert async but in general I have foudn that to be a lil expensive so if we dont need ithis; think its unlikely lets not od it we cand another # invaraint note

attn_gym/linear/kda/api.py
```diff
@@ -33,6 +33,7 @@
 from attn_gym.linear.kda.impl.fused import paged_chunk_forward as _fused_paged_chunk_forward
 from attn_gym.linear.kda.impl.reference import reference_kda
 from attn_gym.linear.kda.naive import naive_chunk_kda, naive_recurrent_kda
+from attn_gym.linear.kda.ops import _ReplayState
```
drisspg, line 36: nit we shoudl jstu define in types and it shoudl be public

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,923 @@
...
+    max_logical_tokens: tl.constexpr,
+    num_heads: tl.constexpr,
+    head_dim: tl.constexpr,
+    is_packed: tl.constexpr,
+    write_output_map: tl.constexpr,
+    USE_INT64_OFFSETS: tl.constexpr,
+    BLOCK: tl.constexpr,
+):
+    sequence = tl.program_id(0)
+    offset = tl.program_id(1) * BLOCK + tl.arange(0, BLOCK)
```
drisspg, line 166:
```suggestion
    sequence = tl.program_id(1)
    offset = tl.program_id(0) * BLOCK + tl.arange(0, BLOCK)
```

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,923 @@
...
+            prefix_caches[:4],
+            tail_caches[:4],
+            strict=True,
+        )
+    ):
+        num_heads, head_dim = source.shape[2:]
+        _pack_prefill_kernel[
+            (
+                num_sequences,
+                triton.cdiv(max_logical_tokens * num_heads * head_dim, pack_block),
```
drisspg, line 735:
```suggestion
                triton.cdiv(max_logical_tokens * num_heads * head_dim, pack_block),
                num_sequences,
```

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,923 @@
...
+            write_output_map=index == 0,
+            BLOCK=pack_block,
+        )
+
+    prefix_beta = prefix_caches[4]
+    tail_beta = tail_caches[4]
+    _pack_prefill_kernel[
+        (
+            num_sequences,
+            triton.cdiv(max_logical_tokens * beta.shape[2], pack_block),
```
drisspg, line 766:
```suggestion
            triton.cdiv(max_logical_tokens * beta.shape[2], pack_block),
            num_sequences,
```

attn_gym/linear/kda/fwd/triton/paged_replay.py
drisspg, line 735: High / must fix — the tile dimension currently lands on grid Y. T=4096, H=32, D=128 produces grid.y=66,544, exceeding CUDA's 65,535 limit, even for dense prefill. The three suggestions move tiles to X: apply both launcher changes (Q/K/V/gate and beta) together with the kernel program-ID change. This is a minimal fix for the reported shape, assuming request count fits grid Y; a flattened 1-D launch avoids that remaining dimension bound. Please add a public replay-prefill regression for this shape. These suggestions are unrun here and do not resolve the separate packed-work scaling issue.

attn_gym/linear/kda/fwd/triton/paged_replay.py
```diff
@@ -0,0 +1,923 @@
...
+    )
+    num_sequences = state_indices.shape[0]
+    prefix_capacity = (
+        triton.cdiv(
+            q.shape[0] * q.shape[1] + (_CHUNK_SIZE - 1) * num_sequences,
+            _CHUNK_SIZE,
+        )
+        * _CHUNK_SIZE
+    )
+    max_logical_tokens = q.shape[1] + _CHUNK_SIZE - 1
```
drisspg, line 695: Must fix — for packed inputs q.shape[1] is the total tokens across all requests, but the launch repeats this bound for each request. For 256 requests × 32 tokens with H=2,D=128, each Q/K/V/gate packing launch still schedules 2,113,280 blocks, mostly masked. Could we mirror the existing ragged persistent scheduling: compute per-request tile counts from pending_count + incoming_length on-device, prefix-sum those counts, then have a bounded worker grid fetch only real tiles? A flattened capacity grid with device-side early exit is also an option; the important part is avoiding the request-count × packed-token-count work space. Simply making the existing Cartesian grid persistent would retain the wasted work. Please measure many-short-sequence prefill and preserve capture-safe metadata handling and exact outputs.

attn_gym/linear/kda/api.py
```diff
@@ -222,6 +224,15 @@ def paged_chunk_kda(
...
             applies to the fused backend as in ``chunk_kda``.
+        replay_state: Optional ``(q, k, v, gate, beta, count)`` persistent cache tuple.
+            Q, K, and V use BF16, gate and beta use FP32, and count uses int32. The five
+            token caches have shape ``[num_slots, 64, ...]``; every count must remain in
+            ``[0, 63]``, and no replay-state tensors may alias. When supplied, calls with one
+            token per dense batch row use replay-backed decode; other calls use streaming
+            prefill. Both paths preserve chunk boundaries across calls and support
+            ``torch.compile`` and CUDA graph capture. Graph replay may change packed offsets,
+            slot routing, initialization flags, and replay counts while tensor shapes remain
+            fixed to the captured bucket.
```
drisspg, line 235: Could we document the boundary-state invariant and memory ownership together here? With replay enabled, state_cache is the FP32 state after the last complete 64-token chunk, while replay_state retains the unfinished suffix. After 150 tokens that means state at 128 plus 22 cached tokens; callers must preserve both across invocations rather than switch to state-only decode and lose the suffix. Also distinguish persistent per-slot replay storage ([num_slots,64,...] for five token caches plus counts) from graph/workspace allocations: the prefix buffers have capacity round_up(total_input_capacity + 63*num_requests,64), and the tail buffers are [num_requests,64,...], in addition to compute intermediates. vLLM needs to budget this per captured bucket; the total additional memory is not just the slot-sized replay cache.

attn_gym/linear/kda/fwd/triton/paged_replay.py
drisspg, line 766: Simplification: could we use one five-tensor loop for Q/K/V/gate/beta here and in commit_paged_chunk_prefill, as prepare_paged_chunk_replay already does? Normalize beta to head_dim=1 with a trailing zero stride, then invoke the same kernel launch/argument block for every tensor. Keep write_output_map/write_count enabled only on the first tensor. The separate beta blocks duplicate most of the launch plumbing and make fixes like the grid-axis change easy to apply to only four of the five caches. This can stay a local loop cleanup rather than adding a new abstraction.

attn_gym/linear/kda/ops.py
```diff
@@ -170,6 +214,36 @@ def _chunk_fwd_ragged_paged_cuda(*args):
...
+
+
+def _chunk_replay_commit_cuda(*args):
+    from attn_gym.linear.kda.fwd.triton.paged_replay import commit_paged_chunk_replay
+
+    return commit_paged_chunk_replay(*args)
+
+
+def _chunk_replay_prefill_prepare_cuda(*args):
+    from attn_gym.linear.kda.fwd.triton.paged_replay import prepare_paged_chunk_prefill
```
drisspg, line 230: why inline improts?

drisspg (approved):
looks good thakns for all the fixes

